PR #1, opened 2018-10-12
add <audio> tags to MSE for audio only

JaKraken:
Add logic to MSE strategy to create <audio> tags for audio only streams, use video for other streams.

Also add unit tests to check that behaviour is similar between video and audio.


script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
-          setUpMSE();
+            expect(videoContainer.childElementCount).toBe(0);
 
-          expect(mseStrategy.transitions.canBePaused()).toBe(true);
-        });
+            mseStrategy.load('src', null, 0);
 
-        it('canBeginSeek() Transition is true', function () {
-          setUpMSE();
+            expect(videoContainer.firstChild).toBe(mockVideoElement);
```
tsadler1988: This test won't fail if we create a video element instead. We need to change the mock to return a mock audio element when`document.createElement` is called with `'audio'`, and assert on that instead.

tsadler1988: Also, this test needs to move into the `load` describe block. It's currently in `transitions` but is not related to transitions.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
-          expect(mockVideoElement.addEventListener).toHaveBeenCalledWith('ended', jasmine.any(Function));
-          expect(mockVideoElement.addEventListener).toHaveBeenCalledWith('error', jasmine.any(Function));
-          expect(mockDashInstance.on).toHaveBeenCalledWith(dashjsMediaPlayerEvents.ERROR, jasmine.any(Function));
-          expect(mockDashInstance.on).toHaveBeenCalledWith(dashjsMediaPlayerEvents.MANIFEST_LOADED, jasmine.any(Function));
-          expect(mockDashInstance.on).toHaveBeenCalledWith(dashjsMediaPlayerEvents.MANIFEST_VALIDITY_CHANGED, jasmine.any(Function));
-          expect(mockDashInstance.on).toHaveBeenCalledWith(dashjsMediaPlayerEvents.QUALITY_CHANGE_RENDERED, jasmine.any(Function));
-          expect(mockDashInstance.on).toHaveBeenCalledWith(dashjsMediaPlayerEvents.METRIC_ADDED, jasmine.any(Function));
-        });
+          it('should attach a new audio source when CDN Failover occurs', function () {
+            setUpMSE(null, null, MediaKinds.AUDIO);
```
tsadler1988: I don't think this test is necessary as the exact same codepath is executed for both audio and video. The only variance is the first assertion should be `initialize` with an audio element, but this is already tested in the above new test.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
+            // Actual live point requested
+            mseStrategy.load('src', null, 0);
 
-          mseStrategy.load('src', null, 0.1);
+            expect(mockDashInstance.initialize).toHaveBeenCalledWith(mockVideoElement, 'src#r=-1', true);
+          });
 
-          expect(mockDashInstance.initialize).toHaveBeenCalledWith(mockVideoElement, 'src#t=101', true);
-        });
+          it('should playback from relative live start time for audio simulcast', function () {
```
tsadler1988: I don't think this test is necessary as the only variance is the initialize should be called with an audio element, but this is already tested in the above new test.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
 
-          mseStrategy.load('src', null, undefined);
+            expect(mockDashInstance.initialize).toHaveBeenCalledWith(mockVideoElement, 'src#r=0', true);
+          });
 
-          expect(mockDashInstance.initialize).toHaveBeenCalledWith(mockVideoElement, 'src', true);
-        });
-      });
+          it('should playback from relative start time for audio simulcast', function () {
+            setUpMSE(0, WindowTypes.SLIDING, MediaKinds.AUDIO);
```
tsadler1988: I don't think this test is necessary as the only variance is the initialize should be called with an audio element, but this is already tested in the above new test.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
-          expect(mockVideoElement.removeEventListener).toHaveBeenCalledWith('ended', jasmine.any(Function));
-          expect(mockVideoElement.removeEventListener).toHaveBeenCalledWith('error', jasmine.any(Function));
-          expect(mockDashInstance.off).toHaveBeenCalledWith(dashjsMediaPlayerEvents.ERROR, jasmine.any(Function));
-          expect(mockDashInstance.off).toHaveBeenCalledWith(dashjsMediaPlayerEvents.QUALITY_CHANGE_RENDERED, jasmine.any(Function));
-          expect(mockDashInstance.off).toHaveBeenCalledWith(dashjsMediaPlayerEvents.METRIC_ADDED, jasmine.any(Function));
-          expect(mockDashInstance.off).toHaveBeenCalledWith(dashjsMediaPlayerEvents.MANIFEST_LOADED, jasmine.any(Function));
-          expect(mockDashInstance.off).toHaveBeenCalledWith(dashjsMediaPlayerEvents.MANIFEST_VALIDITY_CHANGED, jasmine.any(Function));
-        });
+          it('returns the correct duration from the DASH Mediaplayer with audio', function () {
+            setUpMSE(null, null, MediaKinds.AUDIO);
```
tsadler1988: I don't think this test is necessary, as we retrieve the duration from the same dash.js API regardless of whether it's audio or video.

I'm assuming this test was created because the dash.js docs were unclear on duration for audio? I can see why that might prompt you to write a test for it, and it would make sense in an integration test that used dash.js itself, but doesn't make a difference when we've created a mock dash.js that has the same behaviour for `duration()` for both audio and video.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
-          setUpMSE();
-          mseStrategy.load(null, null, 0);
+            mseStrategy.load('src', null, 0);
 
-          expect(videoContainer.childElementCount).toBe(1);
+            expect(mseStrategy.getDuration()).toBe(101);
+          });
 
-          mseStrategy.tearDown();
+          it('returns the correct duration from the DASH Mediaplayer with video', function () {
```
tsadler1988: This seems to be a duplicate of `'returns the correct duration from the DASH Mediaplayer'` further down in the `'getDuration()'` describe block (which is also a more appropriate describe block for this test).

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
+          });
 
-          expect(mockDashInstance.initialize).toHaveBeenCalledWith(mockVideoElement, 'src', true);
-        });
-      });
+          it('should playback from relative start time for audio simulcast', function () {
+            setUpMSE(0, WindowTypes.SLIDING, MediaKinds.AUDIO);
 
-      describe('getSeekableRange()', function () {
-        it('returns the correct start and end time', function () {
```
tsadler1988: As per my above comments re: duration, I don't think it's necessary to test audio and video as the code path is the same.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
+          it('should correctly return the paused state from the MediaPlayer when not paused', function () {
+            setUpMSE();
+            mseStrategy.load(null, null, 0);
+
+            mockDashInstance.isPaused.and.returnValue(false);
+
+            expect(mseStrategy.isPaused()).toBe(false);
+          });
+
+          it('should correctly return the paused state from the audio MediaPlayer when not paused', function () {
```
tsadler1988: As per duration and seekable range above, I don't think we need to test paused for both audio and video.

script/playbackstrategy/msestrategy.js
```diff
@@ -108,8 +109,11 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       }
 
       function setUpMediaElement (playbackElement) {
-        videoElement = document.createElement('video');
-
+        if (mediaKind === MediaKinds.AUDIO) {
+          videoElement = document.createElement('audio');
```
tsadler1988: Can we rename `videoElement` to `mediaElement` or similar? It's a bit misleading otherwise.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,491 +1,579 @@
...
-        mockDashInstance.getDebug.and.returnValue(mockDashDebug);
-      }
+            expect(mseStrategy.transitions.canBePaused()).toBe(true);
+          });
 
-      afterEach(function () {
-        mockVideoElement.currentTime = 0;
-        document.body.removeChild(videoContainer);
-      });
+          it('should create an audio element and add it to the video container', function () {
```
tsadler1988: Can we rename `videoContainer`? It's called `playbackElement` in the implementation, and is no longer just for video as of this change.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -1,579 +1,525 @@
...
+          }
         });
 
-        describe('Load', function () {
-          it('should create a video element and add it to the video container', function () {
-            setUpMSE();
+        window.dashjs = mockDashjs;
 
-            expect(videoContainer.childElementCount).toBe(0);
+        spyOn(document, 'createElement').and.returnValue(mockMediaElement);
```
tsadler1988: Sorry @JaKraken, this is the crucial bit that needs to change to test the correct thing. As it is, calling `createElement` with anything will give you what you're looking for, so the test will always pass. You should be able to see this by reverting your implementation and observing that your test erroneously passes (TDD would've helped you here 😛).

You need to change this mock to behave differently depending on whether it is called with `'video'` or `'audio'` to test the functionality correctly. This can be done with `and.callFake` rather than `and.returnValue`. See https://jasmine.github.io/2.0/introduction.html#section-Spies:_%3Ccode%3Eand.callFake%3C/code%3E.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -106,66 +111,77 @@ require(
...
 
           mseStrategy.load('src', null, 0);
 
-          expect(videoContainer.firstChild).toBe(mockVideoElement);
-          expect(videoContainer.childElementCount).toBe(1);
+          expect(playbackElement.firstChild).toBe(mockMediaElement);
+          expect(playbackElement.childElementCount).toBe(1);
+        });
+
+        it('canBeginSeek() Transition is true', function () {
```
tsadler1988: I think you've accidentally moved this test out of the `'transitions'` describe block.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -417,6 +433,15 @@ require(
           expect(mseStrategy.isPaused()).toBe(false);
         });
 
+        it('should correctly return the paused state from the audio MediaPlayer when not paused', function () {
```
tsadler1988: You've removed most of what I thought was overtesting, but I still think these 2 new audio tests are unnecessary, as this code path is already tested with video in the other tests in this describe block.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -48,16 +48,21 @@ require(
           windowEndTime: windowEndTimeMS || 0
         };
 
-        mockVideoElement = document.createElement('video');
-        mockVideoElement.currentTime = 0;
-        spyOn(mockVideoElement, 'addEventListener');
-        spyOn(mockVideoElement, 'removeEventListener');
+        if (defaultMediaKind === MediaKinds.AUDIO) {
+          mockMediaElement = document.createElement('audio');
```
tsadler1988: Might be better to always create both a video and audio mock element, so when you assert on it you're sure it's the correct one. Then it's up to the mocking of createElement to use the correct one.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -106,66 +111,77 @@ require(
...
+      describe('Load', function () {
+        it('should create a video element and add it to the media element', function () {
           setUpMSE();
 
-          expect(mseStrategy.transitions.canBeginSeek()).toBe(true);
+          expect(playbackElement.childElementCount).toBe(0);
+
+          mseStrategy.load('src', null, 0);
+
+          expect(playbackElement.firstChild).toBe(mockMediaElement);
```
tsadler1988: As per my above comment, it might be better to assert on what you know is an audio element, rather than trust the test setup to have provided the correct element type,

PR #2, opened 2018-10-12
Ensure that NaN is not sent as relative time when playing back simulc…

matt-stephenson:
As the startTime is undefined when attempting to play sliding window content from live, parseInt turns this into NaN and this is used as the relative time on the end of the url passed into dashjs' initialize, e.g.

`http://urltomanifest.mpd#r=NaN`

This small change avoids doing that, in much the same way as we avoid doing it for growing window content a few lines above.


matt-stephenson:
Fixed as part of https://github.com/bbc/bigscreen-player/pull/44

PR #3, opened 2018-10-13
Add Travis build

tsadler1988:
Includes some changes to follow Travis conventions:
 - `npm test` runs the whole build (currently lint and unit tests)
 - `.nvmrc` defines the node version to run the build (because Travis uses `nvm`)


tsadler1988:
Travis is only running 95 tests, not sure why...

matt-stephenson:
> Travis is only running 95 tests, not sure why...

I think the culprit is Squire. There are four test files that don't run at all on Travis, these all have a dependency on Squire, whereas the other files don't.

tsadler1988:
Yep Squire was the problem - https://github.com/bbc/bigscreen-player/commit/3a4a3574beae675e8b1acf3a251244ff4eb5a372.

PR #9, opened 2018-11-09
Clamp seeking time for sliding windows.

tsadler1988:


ShiningTrapez:
Should this change have also changed package-lock.json? Doesn't seem to be any deps updated.

tsadler1988:
@ShiningTrapez if you look at the diff, it's just the bigscreen-player version no. that's changed.

PR #23, opened 2018-12-03
Expose player bitrate

JaKraken:
facilitating reporting of dash playback bitrate and the buffer length through plugins.

Also address some styling issues to the onscreen debug tool.

### DIFF
--- a/script-test/playbackstrategies/msestrategytest.js
+++ b/script-test/playbackstrategies/msestrategytest.js
@@ -15,7 +15,10 @@ require(
       var mockDashDebug;
       var mockVideoElement;
       var eventCallbacks;
+      var dashEventCallback;
       var eventHandlers = {};
+      var mockPlugins;
+      var mockPluginsInterface;
 
       var mockAudioElement = document.createElement('audio');
 
@@ -34,6 +37,14 @@ require(
         injector.mock({
           'dashjs': mockDashjs
         });
+        mockPluginsInterface = jasmine.createSpyObj('interface', ['onErrorCleared', 'onBuffering', 'onBufferingCleared', 'onError', 'onFatalError', 'onErrorHandled', 'onQualityChangeRendered']);
+
+        mockPlugins = {
+          interface: mockPluginsInterface
+        };
+
+        injector.mock({'bigscreenplayer/plugins': mockPlugins});
+
         injector.require(['bigscreenplayer/playbackstrategy/msestrategy'], function (SquiredMSEStrategy) {
           MSEStrategy = SquiredMSEStrategy;
           done();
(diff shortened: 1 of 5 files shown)


script/debugger/debugview.js
```diff
@@ -19,7 +19,6 @@ define('bigscreenplayer/debugger/debugview',
      logBox.style.backgroundColor = '#1D1D1D';
      logBox.style.opacity = 0.9;
      logBox.style.overflow = 'hidden';
-     logBox.style.fontFamily = 'Trebuchet MS';
```
jlks: Seeing as this replicates the functionality in #21 maybe make this monospace just to make the spacing consistent?

Probably best to close that other PR too

script/playbackstrategy/msestrategy.js
```diff
@@ -73,12 +85,30 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
...
-          DebugTool.keyValue({key: 'Dropped Frames', value: event.value.droppedFrames});
+        if (event.mediaType === 'video') {
+          if (event.metric === 'DroppedFrames') {
+            DebugTool.keyValue({key: 'Dropped Frames', value: event.value.droppedFrames});
+          } else if (event.metric === 'BufferLevel') {
+            videoMetrics = mediaPlayer.getMetricsFor('video');
+            dashMetrics = mediaPlayer.getDashMetrics();
+
+            if (videoMetrics && dashMetrics) {
+              playerMetadata.bufferLength = dashMetrics.getCurrentBufferLevel(videoMetrics);
```
jlks: Maybe match the naming up here? buffer length -> buffer level

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -535,5 +547,61 @@ require(
...
+        it('should call plugins with playback bitrate', function () {
+          setUpMSE();
+          mockDashInstance.getBitrateInfoListFor.and.returnValue([{bitrate: 1000}, {bitrate: 2000}, {bitrate: 3000}]);
+          mseStrategy.load(null, null, 0);
+
+          dashEventCallback(dashjsMediaPlayerEvents.QUALITY_CHANGE_RENDERED, mockEvent);
+
+          expect(mockPluginsInterface.onPlayerInfoUpdated).toHaveBeenCalledWith({
+            downloadBitrate: undefined,
+            playbackBitrate: 2,
```
matt-stephenson: Small point, but I had to go on a bit of a journey to see where this number `2` came from. Perhaps a comment could be added, or have the second bitrate something like 2048 and then expect 2.048, so it's easier to make the mental jump.

script/playbackstrategy/msestrategy.js
matt-stephenson: I guess rDot wants bufferLength, so there's a mismatch somewhere. I think here is ok.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -535,5 +547,61 @@ require(
           });
         });
       });
+
+      describe('getMediaPlayerInfo', function () {
```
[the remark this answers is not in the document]
JaKraken: Changed now to reflect new function name, this was due to a refactor

script-test/playbackstrategies/msestrategytest.js
JaKraken: Changed it to your suggestion, hopefully it makes it more easily understood

script/playbackstrategy/msestrategy.js
```diff
@@ -73,12 +84,30 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
...
         }
       }
 
       function onMetricAdded (event) {
-        if (event.mediaType === 'video' && event.metric === 'DroppedFrames') {
-          DebugTool.keyValue({key: 'Dropped Frames', value: event.value.droppedFrames});
+        if (event.mediaType === 'video') {
+          if (event.metric === 'DroppedFrames') {
+            DebugTool.keyValue({key: 'Dropped Frames', value: event.value.droppedFrames});
+          } else if (event.metric === 'BufferLevel') {
```
matt-stephenson: Do we not want to pull this outside of the `if(event.mediaType === 'video')`? Do we get mediaType audio? And if so, we would probably want to call the plugin.

JaKraken: Should now report for audio too, since the code is no longer behind some logic i'm not updating the tests as it should function the same for both audio and video.

PR #35, opened 2019-02-15
Html5 player unit tests

JaKraken:
Adding in the unit tests for the HTML5 media player along with the live modifiers.

If you spot any way these can be improved please do let me know!


script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+              expect(mockVideoMediaElement.currentTime).toEqual(10);
+            });
+
+            it(' Get Player Element Returns Video Element For Video', function () {
+              player.initialiseMedia(MediaPlayerBase.TYPE.VIDEO, 'http://url/', 'video/mp4', sourceContainer, {});
+
+              expect(player.getPlayerElement()).toEqual(mockVideoMediaElement);
+            });
+
+            it(' Get Player Element Returns Audio Element For Audio', functi
```
magnuswoody: Could `playFromCurrentTime` be used for the test above as well? It seems `playFromCurrentTime` is only used once

magnuswoody: ambiguous? maybe use the normal Jasmine clock method to be clear on what it's doing

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+
+        var recentEvents = [];
+
+        var logger = jasmine.createSpyObj('logger', ['warn', 'debug', 'error']);
+
+        mockSourceElement = document.createElement('source');
+        mockVideoMediaElement = document.createElement('video');
+        mockAudioMediaElement = document.createElement('audio');
+
+        function playFromNearCurrentTime (currentTime, targetTime) {
```
tsadler1988: This function is only used once - should it just be part of the test that uses it instead?

script-test/playbackstrategies/modifiers/html5tests.js
magnuswoody: Should these be converted?

magnuswoody: Use of `self` intended?

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+          logger = jasmine.createSpyObj('logger', ['warn', 'debug', 'error']);
+          sourceContainer = document.createElement('div');
+
+          injector.require(['bigscreenplayer/playbackstrategy/modifiers/html5'], function (SquiredHtml5) {
+            html5Player = SquiredHtml5;
+            done();
+          });
+        });
+
+        describe('tests', function () {
```
tsadler1988: Unnecessary `describe` block. If removed, you could also combine the `beforeEach`s.

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+
+            it(' Pause Not Passed From Media Element To Media Player On User Pause From Buffering', function () {
+              player.initialiseMedia(MediaPlayerBase.TYPE.VIDEO, 'http://url/', 'video/mp4', sourceContainer, {});
+
+            // We dont fire the metadata ready callback so it stays in the buffering state
+              player.beginPlayback();
+
+              mockVideoMediaElement.pause();
+
+              expect(player.toPaused).toHaveBeenCalledTimes(0);
```
tsadler1988: `toPaused` should not be part of the public API, and testing that the implementation calls it is a bit white box-y. I'd suggest you assert of the effects of `toPaused` instead (event sent, state changed).

Same goes for other usages below. Also means you can stop spying on it in the `beforeEach`.

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+              }
+            });
+
+            spyOn(mockVideoMediaElement, 'play');
+            spyOn(mockVideoMediaElement, 'load');
+            spyOn(mockVideoMediaElement, 'pause');
+          });
+
+          afterEach(function () {
+            player = null;
```
tsadler1988: Not sure this is necessary as it gets re-instantiated in `beforeEach`.

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+
+              player.reset();
+
+              expect(player.getDuration()).toBe(undefined);
+            });
+
+            it('Calling Begin Playback In Empty State Is An Error', function () {
+              player.beginPlayback();
+
+              expect(logger.error).toHaveBeenCalledWith('Cannot beginPlayback while in the \'EMPTY\' state');
```
tsadler1988: You might want to assert on more than just the logger - when these types of error happen (bad API calls), we also change state to error, fire an event, and essentially reset the media player.

Same goes for all tests like this.

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+              expect(player.getState()).toEqual(MediaPlayerBase.STATE.ERROR);
+
+              metaDataCallback();
+              finishedBufferingCallback();
+
+              expect(player.getState()).toEqual(MediaPlayerBase.STATE.ERROR);
+            });
+          });
+
+          describe('Error logged for invalid state transitions tests', function () {
```
tsadler1988: Could possibly fold these into other error state tests.

script-test/playbackstrategies/modifiers/html5tests.js
tsadler1988: This and the next few tests seem to be doing too much. The point of the these tests is that we don't actually perform a seek, and our player code feigns a transition from buffering and then back to playing, as if we had seeked. But these tests emulate real buffering instead.

script-test/playbackstrategies/modifiers/html5tests.js
JaKraken: These are testing calls on methods when in states that are incorrect, the other is testing functionality within only the error state.

script-test/playbackstrategies/modifiers/html5tests.js
```diff
@@ -0,0 +1,3260 @@
...
+
+        function giveMediaElementMetaData (mediaElement, metadata) {
+          try {
+            spyOnProperty(mediaElement, 'seekable').and.returnValue(
+              {
+                start: function () {
+                  return metadata.start;
+                },
+                end: function () {
+                  return metadata.end;
```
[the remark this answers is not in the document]
JaKraken: I think it will do for now unless anyone else raises some objection to this being used

script-test/playbackstrategies/modifiers/live/restartabletest.js
```diff
@@ -0,0 +1,330 @@
...
+              });
+
+              initialiseRestartableMediaPlayer();
+            });
+
+            afterEach(function () {
+              jasmine.clock().uninstall();
+              mockCallback = [];
+            });
+              // Buffering and time out tests here
```
magnuswoody: Comment left on purpose?

script-test/playbackstrategies/modifiers/live/seekabletest.js
```diff
@@ -0,0 +1,346 @@
...
+            player.addEventCallback.and.callFake(function (self, callback) {
+              mockCallback.push(callback);
+            });
+          });
+
+          afterEach(function () {
+            jasmine.clock().uninstall();
+            mockCallback = [];
+          });
+            // Buffering and time out tests here
```
magnuswoody: See previous comment above in `restartabletest.js` -> should comments be left?

script/playbackstrategy/modifiers/html5.js
```diff
@@ -522,7 +522,7 @@ define(
         var clampedTime = getClampedTime(seconds);
         var range = getSeekableRange();
         if (clampedTime !== seconds) {
-          logger.debug('playFrom ' + seconds + ' clamped to ' + clampedTime + ' - seekable range is { start: ' + range.start + ', end: ' + range.end + ' }');
```
magnuswoody: Any reason why this changed?

JaKraken: Matches up with tests, changed this one because its quicker

PR #39, opened 2019-03-05
Logging Levels

jlks:
An initial implementation of logging levels. This also fixes the issue with displaying the logs when they are initially toggled to show up on screen. Raising a PR for visibility and feedback.


script/debugger/debugtool.js
```diff
@@ -11,6 +11,15 @@ define('bigscreenplayer/debugger/debugtool',
...
      var visible = false;
 
+     var LOG_LEVELS = {
+       ERROR: 0,
+       DEBUG: 1,
+       INFO: 2,
+       VERBOSE: 3
+     };
+
+     var logLevel = LOG_LEVELS.VERBOSE;
```
magnuswoody: Do we want this verbose by default? Won't this include all of the Dash.js debug when we are just using the debug tool?

PR #44, opened 2019-04-15
Live cdn failover

matt-stephenson:
Failover for LIVE playback - DASH only.

- Native Strategy - uses the `load` functions `startTime` argument as the time to failover to. This is passed in by `playerComponent` using `getCurrentTime`, which is the native strategy's internally stored current time.
- MSE Strategy - ignores the `load` functions `playbackTime` argument and instead uses the MSE strategy's internally stored `failoverTime`.

Refactoring unit tests.

### DIFF
--- a/script-test/playercomponenttest.js
+++ b/script-test/playercomponenttest.js
@@ -951,23 +951,6 @@ require(
           expect(mockStateUpdateCallback.calls.mostRecent().args[0].data.state).toEqual(MediaState.FATAL_ERROR);
         });
 
-        // it('should fire a fatal error on the plugins if the playback type is live', function () {
-        //   spyOn(mockStrategy, 'getDuration').and.returnValue(100);
-        //   spyOn(mockStrategy, 'getCurrentTime').and.returnValue(94);
-        //   spyOn(mockStrategy, 'getSeekableRange').and.returnValue({start: 0, end: 100});
-        //   spyOn(mockStrategy, 'load');
-
-        //   setUpPlayerComponent({windowType: WindowTypes.SLIDING, multiCdn: true});
-
-        //   mockStrategy.mockingHooks.fireErrorEvent({errorProperties: {}});
-
-        //   jasmine.clock().tick(5000);
-
-        //   expect(mockStrategy.load).toHaveBeenCalledTimes(1);
-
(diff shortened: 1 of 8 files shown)


script-test/playercomponenttest.js
```diff
@@ -950,37 +951,32 @@ require(
           expect(mockStateUpdateCallback.calls.mostRecent().args[0].data.state).toEqual(MediaState.FATAL_ERROR);
         });
 
-        it('should fire a fatal error on the plugins if the playback type is live', function () {
```
tsadler1988: Feel like these 2 deleted tests should still exist for:
 - Non-dash
 - Fatal error when not remaining CDN's to failover to

script/playbackstrategy/msestrategy.js
```diff
@@ -67,10 +65,15 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       }
 
       function onTimeUpdate () {
+        failoverTime = mediaElement.currentTime - timeCorrection;
         publishTimeUpdate();
       }
 
       function onError (event) {
+        if (event.error && event.error.data) {
```
tsadler1988: Can you elaborate on this? I've seen the commit message 'remove error data we do not need for now' but more context would be useful.

script-test/playercomponenttest.js
matt-stephenson: Yep good call. We've added an explicit test for not failing over when playing back HLS content.

For the no remaining CDN's, rather than a test getting us failing over multiple times, we've added extra assertions that the CDN url list is being sliced, and there is already a test that we throw fatal error if we only have one CDN, so this should be covered.

script/playbackstrategy/msestrategy.js
matt-stephenson: I'm assuming this is a hygiene thing that was spotted, for the extra data dashjs sends through on its error event (vs the one from the media element), and not directly related to CDN failover? Is this correct @jlks?

jlks: Yep, that is correct. 

There is a deep clone of the error event object that happens as a part of the plugin code path that can't handle some of the nested objects that are returned from that data object (Request/Response etc).

PR #47, opened 2019-04-18
Webcast Failover for Seekable HLS Devices

ShiningTrapez:


script-test/playercomponenttest.js
```diff
@@ -974,6 +976,54 @@ require(
           expect(mockStateUpdateCallback.calls.mostRecent().args[0].data.state).toEqual(MediaState.FATAL_ERROR);
         });
 
+        it('should publish a media state update of fatal if playback is live hls webcast with playable', function () {
+          spyOn(mockStrategy, 'getDuration').and.returnValue(100);
+          spyOn(mockStrategy, 'getCurrentTime').and.returnValue(94);
+          spyOn(mockStrategy, 'load');
+
+          setUpPlayerComponent({multiCdn: true, manifestType: 'm3u8', windowType: WindowTypes.GROWING, liveSupport: 'playable'});
```
tsadler1988: Could use the live support enum here.

script-test/playercomponenttest.js
```diff
@@ -974,6 +976,54 @@ require(
...
+
+          expect(mockStrategy.load).toHaveBeenCalledTimes(1);
+
+          expect(mockStateUpdateCallback.calls.mostRecent().args[0].data.state).toEqual(MediaState.FATAL_ERROR);
+        });
+
+        it('should cdn failover on hls webcast with seekable', function () {
+          var secondCdn = 'b';
+          var currentTime = 10;
+          var type = 'application/dash+xml';
```
tsadler1988: Using DASH mime type here is confusing. I suspect you'll have to change the stub data in `setUpPlayerComponent` for this test to change it to HLS.

script/playercomponent.js
```diff
@@ -214,7 +216,8 @@ define(
         var hasNextCDN = mediaMetaData.urls.length > 1;
         var aboutToEndVod = getDuration() > 0 && (getDuration() - getCurrentTime()) <= 5;
         var canVodFailover = windowType === WindowTypes.STATIC && !aboutToEndVod;
-        var canLiveFailover = windowType !== WindowTypes.STATIC && transferFormat === 'dash';
+        var hlsSupported = windowType === WindowTypes.GROWING && liveSupport === LiveSupport.SEEKABLE;
```
tsadler1988: Can I be super picky and ask for this to renamed? This reads to me like it's referring to device support, rather than information on the current media. Perhaps `canHlsLiveFailover`? Could maybe extract a `canDashLiveFailover` as well.

script/livesupportenum.js
```diff
@@ -0,0 +1,15 @@
+define(
+  'bigscreenplayer/livesupportenum',
```
tsadler1988: Can this be moved into `script/models` with the other BSP enums?

script-test/playercomponenttest.js
ShiningTrapez: Do we use `application/x-mpegURL`, or `application/vnd.apple.mpegurl`?

tsadler1988: `application/vnd.apple.mpegurl`

script/playercomponent.js
```diff
@@ -214,9 +216,10 @@ define(
         var hasNextCDN = mediaMetaData.urls.length > 1;
         var aboutToEndVod = getDuration() > 0 && (getDuration() - getCurrentTime()) <= 5;
         var canVodFailover = windowType === WindowTypes.STATIC && !aboutToEndVod;
-        var canLiveFailover = windowType !== WindowTypes.STATIC && transferFormat === 'dash';
+        var canHlsLiveFailover = windowType === WindowTypes.GROWING && liveSupport === LiveSupport.SEEKABLE;
```
Luke-Chatburn: This doesn't look correct to me, this should be sliding window?

Luke-Chatburn: Do we need to check for transfer format here as well? because this will come back true for dash

ShiningTrapez: It's probably worth noting as future changes could break this but currently, if `windowType === GROWING`, then `windowType !== STATIC` and it would pass the DASH Check anyway

PR #48, opened 2019-04-18
Live CDN failover time fix

jlks:
## Impact

Sliding window CDN failover using dash.js was previously using the time correction based on the initial stream start time to determine the time to restart. 

This causes the time that CDN failover occurs to drift by the amount of time you have been watching a live stream.

It was also possible to failover to just behind the live point when watching from the start of a window, so this has been clamped to 0 for this case. This was due to #r= using negative numbers, IN_STREAM_BUFFERING being 20 seconds and the `failoverTime` being < 20.

### DIFF
--- a/script/playbackstrategy/msestrategy.js
+++ b/script/playbackstrategy/msestrategy.js
@@ -65,7 +65,14 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       }
 
       function onTimeUpdate () {
-        failoverTime = mediaElement.currentTime - timeCorrection;
+        var IN_STREAM_BUFFERING_SECONDS = 20;
+        var dvrInfo = mediaPlayer.getDashMetrics().getCurrentDVRInfo(mediaPlayer.getMetricsFor('video'));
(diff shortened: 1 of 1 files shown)


script/playbackstrategy/msestrategy.js
```diff
@@ -65,7 +65,14 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       }
 
       function onTimeUpdate () {
-        failoverTime = mediaElement.currentTime - timeCorrection;
+        var IN_STREAM_BUFFERING_SECONDS = 20;
+        var dvrInfo = mediaPlayer.getDashMetrics().getCurrentDVRInfo(mediaPlayer.getMetricsFor('video'));
+        failoverTime = mediaElement.currentTime;
```
tsadler1988: Super picky, but this would be more readable in an `else`. Reading it made me as 'why would `SLIDING` window ever use this?'

But given we plan to redo MSE failover in the near future, happy to approve this for now.

PR #57, opened 2019-05-01
Load manifest

matt-stephenson:
Breaking API changes.

bigscreen-player initialisation is now asynchronous. 

1. the manifest is no longer passed in as part of `bigscreenPlayerData`. Rather, if needed (live playback and not a `playable` device), the manifest will be loaded and parsed in order to obtain the `time` block (windowStartTime, windowEndTime, timeCorrection).

2. `liveSupport` no longer needs to be passed into the initialisation. This is reasoned from the strategies, and also exposed on the `bigscreen-player` api as a static function.

3. a `callbacks` object is required, with `onSuccess` and `onError`. These are called on successful initialisation / failure to load manifest respectively.

Note:
- LoadUrl code is a duplication of the TAL function, which we will accept for now, but will be revisited.
- New enums for transfer formats and live support
- Sinon added as a dev dependency (for custom responses to network requests in unit tests)


script/bigscreenplayer.js
```diff
@@ -82,58 +82,77 @@ define('bigscreenplayer/bigscreenplayer',
...
+          Chronicle.init();
+          device = newDevice;
+          windowType = newWindowType;
+          serverDate = bigscreenPlayerData.serverDate;
 
-          DebugTool.keyValue({key: 'available cdns', value: availableCdns});
-          DebugTool.keyValue({key: 'current cdn', value: bigscreenPlayerData.media.urls[0].cdn});
-          DebugTool.keyValue({key: 'url', value: bigscreenPlayerData.media.urls[0].url});
+          if (needToGetManifest() && !bigscreenPlayerData.time) {
+            ManifestLoader().load(
```
magnuswoody: Should this be a function or reference a specific instance? Rather than instantiating it everytime

script/manifest/manifestparser.js
```diff
@@ -33,7 +33,7 @@ define('bigscreenplayer/parsers/manifestparser',
          return {
            windowStartTime: windowStartTime,
            windowEndTime: windowEndTime,
-           timeCorrection: timeCorrection
+           correction: timeCorrection
```
magnuswoody: Any reason for the name change?

script/playercomponent.js
```diff
@@ -216,8 +216,8 @@ define(
         var hasNextCDN = mediaMetaData.urls.length > 1;
         var aboutToEndVod = getDuration() > 0 && (getDuration() - getCurrentTime()) <= 5;
         var canVodFailover = windowType === WindowTypes.STATIC && !aboutToEndVod;
-        var canHlsLiveFailover = windowType === WindowTypes.GROWING && (liveSupport === LiveSupport.SEEKABLE || liveSupport === LiveSupport.PLAYABLE) && transferFormat === 'hls';
-        var canDashLiveFailover = windowType !== WindowTypes.STATIC && transferFormat === 'dash';
+        var canHlsLiveFailover = windowType === WindowTypes.GROWING && (getLiveSupport() === LiveSupport.SEEKABLE || getLiveSupport() === LiveSupport.PLAYABLE) && transferFormat === TransferFormats.HLS;
```
RonanBBC: These `getLiveSupport` calls need device passed in in case we're using tal strategy

matt-stephenson: 👍 good spot

script/manifest/manifestparser.js
matt-stephenson: The strategies do need this to be `correction`. If you refer to the original bigscreenplayer.js line 105, you can see where the return block from this parser is used - timeCorrection is written into time.correction.

matt-stephenson: https://github.com/bbc/bigscreen-player/pull/57/files/a253f11a91b152329b1923f9991409757430a524#diff-d5e95b3ca57b1de771e8a60bafec18b9L105

script/bigscreenplayer.js
```diff
@@ -82,58 +82,77 @@ define('bigscreenplayer/bigscreenplayer',
         return windowStartTime ? windowStartTime + (seconds * 1000) : undefined;
       }
 
-      return {
-        init: function (playbackElement, bigscreenPlayerData, newWindowType, enableSubtitles, newLiveSupport, device) {
-          Chronicle.init();
+      // Will need reusing in Player Component for Live CDN failover HLS
```
jlks: Leave a TODO here at least if this needs to be done

matt-stephenson: 👍

script/playercomponent.js
```diff
@@ -392,5 +394,13 @@ define(
         tearDown: tearDown
       };
     };
+
+    function getLiveSupport (device) {
+      return PlaybackStrategy.getLiveSupport(device);
+    }
+
+    PlayerComponent.getLiveSupport = getLiveSupport;
```
Luke-Chatburn: Can this not be added on to the object like all the other public functions?

RonanBBC: No because tap-static needs to use it before Bigscreen Player is initialised.

PR #58, opened 2019-05-07
DASH baseurl failover

jlks:
The DVB DASH spec allows for multiple baseurls to be defined in the MPD that are then used for playback failover. 

This PR combines the urls passed in and adds them as baseurls to a manifest.

Note: this does rely on the same path layout for media across the CDNs


script/playbackstrategy/msestrategy.js
```diff
@@ -120,6 +141,55 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         }
       }
 
+      function createPlaybackProperties () {
```
RonanBBC: This would read a lot clearer if you returned an object literal rather than created an empty object then adding the keys

script/playercomponent.js
```diff
@@ -178,7 +180,7 @@ define(
         bubbleBufferingCleared(playbackProperties);
 
         var playbackErrorProperties = createPlaybackErrorProperties(event);
-        raiseError(playbackErrorProperties);
+        raiseError(playbackErrorProperties, false); // TODO: Why was this never set before?
```
RonanBBC: Any unset parameters are passed as `undefined` so would evaluate as falsey anyway

script/playbackstrategy/msestrategy.js
```diff
@@ -120,6 +141,57 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
...
+        playbackProperties.seekable_range = getSeekableRange().start + ' to ' + getSeekableRange().end;
+        playbackProperties.current_time = getCurrentTime();
+        playbackProperties.duration = getDuration();
+
+        return playbackProperties;
+      }
+
+      function propagateCdnFailover (event, cdn) {
+        // Initial playback
+        if (cdn === mediaSources[0].cdn) return;
```
matt-stephenson: Should we get into this code on initial playback? It looks like we only get here when dashjs throws a CDN_FAILOVER event.

script/playercomponent.js
matt-stephenson: I would vote to keep the parameter for clarity, but remove the TODO.

script/playbackstrategy/msestrategy.js
jlks: onCdnFailover gets called every time there is a BASEURL_SELECTED event fired from dashjs. I would have preferred to use SERVICE_LOCATION_BLACKLIST_ADD or SERVICE_LOCATION_BLACKLIST_CHANGED but then we don't get the baseurl back to match up with our other information

script/debugger/cdndebugoutput.js
```diff
@@ -0,0 +1,43 @@
...
+        return element && element.cdn;
+      });
+    }
+
+    function updateMedia () {
+      media.shift();
+      currentCDN = media[0];
+    }
+
+    function update (cdn) {
```
matt-stephenson: Looks like `cdn` argument is unused here

script/playbackstrategy/msestrategy.js
```diff
@@ -121,6 +142,57 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
...
+          stateType: PluginEnums.TYPE.ERROR,
+          properties: errorProperties,
+          isBufferingTimeoutError: false,
+          cdn: cdn
+        });
+        // urls -> sources -> mediaSources
+        // TODO: Remove this horrible mutation when failover is pushed down per strategy.
+        mediaSources.shift();
+        Plugins.interface.onErrorHandled(evt);
+        cdnDebugOutput.update(cdn);
```
matt-stephenson: Can remove argument.

script/playbackstrategy/msestrategy.js
```diff
@@ -121,6 +142,57 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
...
+
+        var errorProperties = PlaybackUtils.merge(createPlaybackProperties(), event.errorProperties);
+        var evt = new PluginData({
+          status: PluginEnums.STATUS.FAILOVER,
+          stateType: PluginEnums.TYPE.ERROR,
+          properties: errorProperties,
+          isBufferingTimeoutError: false,
+          cdn: cdn
+        });
+        // urls -> sources -> mediaSources
```
matt-stephenson: Could we make this comment more the _why_ than the _what_?

script/playbackstrategy/msestrategy.js
```diff
@@ -42,6 +42,7 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       var DashJSEvents = {
         ERROR: 'error',
         MANIFEST_LOADED: 'manifestLoaded',
+        DOWNLOD_MANIFEST_ERROR_CODE: 26,
```
matt-stephenson: Typo - DOWNLOAD. Sorry!

PR #71, opened 2019-06-13
Fix mse base urls for absolute urls

RonanBBC:


script/manifest/manifestmodifier.js
```diff
@@ -0,0 +1,59 @@
...
+
+    function extractBaseUrl (manifest) {
+      return manifest.Period && manifest.Period.BaseURL || manifest.BaseURL && manifest.BaseURL.__text;
+    }
+
+    function generateBaseUrls (manifest, sources) {
+      var baseUrl = extractBaseUrl(manifest);
+      if (!baseUrl) return;
+      if (baseUrl.match(/^https?:\/\//)) {
+        sources = sources.slice(0, 1);
```
RonanBBC: We need to modify the manifest to have the correct `serviceLocation` even though the url is not changing. This is so we know which events to ignore in `msestrategy -> propagateCdnFailover`

script-test/manifest/manifestmodifiertest.js
```diff
@@ -165,11 +165,92 @@ require(
...
+            }
+          };
+
+          expect(ManifestModifier.extractBaseUrl(manifest)).toBe('dash/');
+        });
+
+        it('should return the base url from the root', function () {
+          var manifest = {
+            BaseURL: {
+              __text: 'https://cnd/dash/'
```
matt-stephenson: CND? :)

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -407,9 +413,9 @@ require(
 
         it('should call plugin handler on dash baseUrl changed event', function () {
```
matt-stephenson: Could this test name have `manifest loaded event` rather than `baseurl changed event`

PR #77, opened 2019-06-26
Teardown seeking for restartable devices

RonanBBC:
We can simulate seeking by creating a fake version of the current time (and seekable range), and then restarting (complete teardown) the media player to begin playback from the user's input seeking time. This will enable `restartable` devices with unreliable `mediaElement.currentTime` to now have seeking capabilities.

### DIFF
--- a/script-test/playbackstrategies/modifiers/live/restartabletest.js
+++ b/script-test/playbackstrategies/modifiers/live/restartabletest.js
@@ -104,6 +104,8 @@ require(
             });
 
             it('calls pause on the media player', function () {
+              player.getSeekableRange.and.returnValue({start: 0});
+
               wrapperTests('pause');
             });
 
@@ -184,6 +186,7 @@ require(
             var mockCallback = [];
 
             function startPlaybackAndPause (startTime, disableAutoResume) {
+              player.getCurrentTime.and.returnValue(startTime);
               restartableMediaPlayer.beginPlaybackFrom(startTime);
               restartableMediaPlayer.pause({disableAutoResume: disableAutoResume});
             }
@@ -195,6 +198,7 @@ require(
               player.addEventCallback.and.callFake(function (self, callback) {
                 mockCallback.push(callback);
               });
+              player.getSeekableRange.and.returnValue({start: 0});
 
               initialiseRestartableMediaPlayer();
             });
@@ -267,6 +271,7 @@ require(
               startPlaybackAndPause(20, false);
 
               jasmine.clock().tick(12 * 1000);
+              player.getSeekableRange.and.returnValue({start: 12});
 
               restartableMediaPlayer.pause();
 
--- a/script-test/playbackstrategies/modifiers/live/restartabletest.js
+++ b/script-test/playbackstrategies/modifiers/live/restartabletest.js
@@ -32,7 +32,7 @@ require(
             player = jasmine.createSpyObj('player',
               ['beginPlayback', 'initialiseMedia', 'stop', 'reset', 'getState', 'getSource', 'getMimeType',
                 'addEventCallback', 'removeEventCallback', 'removeAllEventCallbacks', 'getPlayerElement', 'pause',
-                'resume', 'beginPlaybackFrom', 'getCurrentTime']);
+                'resume', 'beginPlaybackFrom', 'getCurrentTime', 'getSeekableRange']);
 
             function mockMediaPlayer () {
               return player;
@@ -106,6 +106,14 @@ require(
             it('calls pause on the media player', function () {
               wrapperTests('pause');
             });
+
+            it('getCurrentTime', function () {
+              wrapperTests('getCurrentTime');
+            });
+
+            it('getSeekableRange', function () {
+              wrapperTests('getSeekableRange');
+            });
           });
 
           describe('should not have methods for', function () {
@@ -119,14 +127,6 @@ require(
 
             it('playFrom', function () {
               isUndefined('playFrom');
-            });
-
-            it('getCurrentTime', function () {
-              isUndefined('getCurrentTime');
-            });
-
-            it('getSeekableRange', function () {
-              isUndefined('getSeekableRange');
             });
           });
 
--- a/script-test/playbackstrategies/modifiers/live/restartabletest.js
+++ b/script-test/playbackstrategies/modifiers/live/restartabletest.js
@@ -11,7 +11,7 @@ require(
 
         function initialiseRestartableMediaPlayer (config, windowType) {
           windowType = windowType || WindowTypes.SLIDING;
-          restartableMediaPlayer = RestartableMediaPlayer(player, config, windowType, { windowStartTime: 0, windowEndTime: 100000 });
+          restartableMediaPlayer = RestartableMediaPlayer(player, config, windowType);
         }
 
         describe('restartable HMTL5 Live Player', function () {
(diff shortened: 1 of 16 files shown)


script/playbackstrategy/modifiers/live/restartable.js
```diff
@@ -6,49 +6,25 @@ define(
...
     function (Html5Player, MediaPlayerBase) {
       'use strict';
-      var AUTO_RESUME_WINDOW_START_CUSHION_MILLISECONDS = 8000;
+      var AUTO_RESUME_WINDOW_START_CUSHION_SECONDS = 8;
 
-      function RestartableLivePlayer (deviceConfig, logger) {
-        var mediaPlayer = Html5Player(logger);
-        var millisecondsUntilStartOfWindow;
-        var bufferingStarted;
+      function RestartableLivePlayer (deviceConfig, logger, windowType, timeData) {
```
jlks: Could always pass in a mediaPlayer instance into this instead of instantiating it... Also stops the implementation being tied to HTML5Player

script/playbackstrategy/modifiers/html5.js
```diff
@@ -35,6 +36,16 @@ define(
       var sentinelIntervalNumber;
       var lastSentinelTime;
 
+      var fakeTimer = {
+        currentTime: undefined,
+        runningTime: undefined,
+        window: {
```
jlks: Would seekableRange be a more appropriate property name? It would be more consistent with the other usages

magnuswoody: Pulled out and moved into `restartable.js`.

script/playbackstrategy/modifiers/live/restartable.js
magnuswoody: All live modifiers now require the media player passed in.

script/playbackstrategy/modifiers/live/restartable.js
```diff
@@ -96,23 +80,37 @@ define(
           }
         }
 
+        function getCurrentTime () {
+          return fakeTimer.currentTime;
+        }
+
+        function getSeekableRange () {
+          var windowLength = (timeData.windowEndTime - timeData.windowStartTime) / 1000;
```
tsadler1988: As discussed, `restartable` devices should report a reliable seekable range when it first starts playing, before it becomes inaccurate over time. This will generally be shorter than the window, as browsers often struggle seeking to the extremities of the window. 

As such, it will be more correct to the initially reported seekable range from the browser a a baseline, rather than the extremities of the window.

script/playbackstrategy/modifiers/live/restartable.js
```diff
@@ -1,87 +1,71 @@
...
+          function newCallback (event) {
+            updateFakeTimer(event.state);
 
-        function removeEventCallback (thisArg, callback) {
-          mediaPlayer.removeEventCallback(thisArg, callback);
+            event.currentTime = getCurrentTime();
+            event.seekableRange = getSeekableRange();
+            callback(event);
+          }
+          callbacksMap.push({ from: callback, to: newCallback });
```
tsadler1988: If/when we fix the problem with the underlying player's event callbacks, we might want to extract this functionality and re-use. Should probably raise an issue for that actually...

script/playercomponent.js
```diff
@@ -126,10 +127,31 @@ define(
       function setCurrentTime (time) {
         userInteracted = true;
         if (transitions().canBeginSeek()) {
-          playbackStrategy.setCurrentTime(time);
+          if (windowType !== WindowTypes.STATIC && getLiveSupport(device) === LiveSupport.RESTARTABLE) {
```
tsadler1988: Per my above comment in `dynamicwindowutils.js`, this should only apply for native strategy, not TAL.

script/playercomponent.js
```diff
@@ -126,10 +127,31 @@ define(
...
+            playbackStrategy.setCurrentTime(time);
+          }
         }
       }
 
+      function reloadMediaElement (time) {
+        function doSeek (time) {
+          var thenPause = playbackStrategy.isPaused();
+          tearDownMediaElement();
+          if (time > (bigscreenPlayerData.time.windowEndTime - bigscreenPlayerData.time.windowStartTime) / 1000 - 10) {
```
tsadler1988: Assuming you will revisit this depending on whether my suggestion for seekable range works? You might want to use the `playbackStrategy.getSeekableRange()` to determine whether or not to set a start time. And you might not want to give it an arbitrary 10 second cushion.

script/playercomponent.js
```diff
@@ -228,37 +250,41 @@ define(
...
-          {
-            onSuccess: function (manifestData) {
-              var windowOffset = manifestData.time.windowStartTime - getWindowStartTime();
-              bigscreenPlayerData.time = manifestData.time;
-              failoverTime -= windowOffset / 1000;
-              cdnFailover(failoverTime, thenPause, errorProperties, bufferingTimeoutError);
-            },
-            onError: function () {
-              bubbleFatalError(errorProperties, bufferingTimeoutError);
+      function reloadManifest (time, successCallback, errorCallback) {
```
tsadler1988: Can we either rename this, or move the `if` logic out to usages of it? It's a bit misleading to call it `reloadManifest` when it might not.

script/playercomponent.js
```diff
@@ -228,37 +250,41 @@ define(
...
+        if (transferFormat === TransferFormats.HLS && LiveSupportUtils.needToGetManifest(windowType, getLiveSupport(device))) {
+          ManifestLoader.load(
+            bigscreenPlayerData.media.urls,
+            bigscreenPlayerData.serverDate,
+            {
+              onSuccess: function (manifestData) {
+                var windowOffset = manifestData.time.windowStartTime - getWindowStartTime();
+                bigscreenPlayerData.time = manifestData.time;
```
tsadler1988: An existing bug, but we really need to propagate the updated time data to the playback strategy. I think @jlks and @matt-stephenson are looking into this so maybe we can live with it for this PR if it doesn't break fake seeking?

spec/support/boot.js
```diff
@@ -74,7 +74,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
   var filterSpecs = !!queryString.getParam('spec');
 
   // Overrides
-  env.stopOnSpecFailure(true);
+  // env.stopOnSpecFailure(true);
```
magnuswoody: Executive decision to leave this out?

script-test/playbackstrategies/modifiers/live/restartabletest.js
```diff
@@ -120,13 +109,93 @@ require(
...
+
+                timeUpdate({ state: MediaPlayerBase.STATE.PLAYING });
+
+                expect(restartableMediaPlayer.getCurrentTime()).toBe(10);
+              });
             });
 
-            it('getSeekableRange', function () {
-              isUndefined('getSeekableRange');
+            describe('getSeekableRange', function () {
```
matt-stephenson: Should there be a test in here for the seekable range when calling `beginPlayback` (as opposed to beginPlaybackFrom`), as there is in the currentTime describe block above?

script-test/playbackstrategies/modifiers/live/restartabletest.js
```diff
@@ -120,13 +109,93 @@ require(
...
+              it('should start at the window time', function () {
+                restartableMediaPlayer.beginPlaybackFrom(0);
+
+                timeUpdate({ state: MediaPlayerBase.STATE.PLAYING });
+
+                expect(restartableMediaPlayer.getSeekableRange()).toEqual({ start: 0, end: 100 });
+              });
+
+              it('should increase with time', function () {
```
matt-stephenson: Could the test name explicitly call out that this is for sliding window (as the test below calls it out for growing)

script-test/playbackstrategies/modifiers/live/restartabletest.js
RonanBBC: The code for `beginPlayback` and `beginPlaybackFrom` is identical for seekable range.

spec/support/boot.js
RonanBBC: Overruled

script/playercomponent.js
RonanBBC: I don't think you should care if you need to reload the manifest or not so I'd prefer the if statement to be kept inside.

PR #81, opened 2019-07-15
Failover on Native-Restartable devices

magnuswoody:
Using the `restartable` fake time functionality, we can now allow live CDN failover on `nativestrategy`, `restartable` devices. As part of this work, a configurable window flag (`window.bigscreenPlayer.disableLiveFailover`) can be set to disable live failover for any specific conditions.

### DIFF
--- a/script/mediaresilience.js
+++ b/script/mediaresilience.js
@@ -1,17 +1,15 @@
 define(
   'bigscreenplayer/mediaresilience', [
     'bigscreenplayer/models/windowtypes',
-    'bigscreenplayer/models/playbackstrategy',
-    'bigscreenplayer/models/livesupport',
-    'bigscreenplayer/models/transferformats'
+    'bigscreenplayer/models/playbackstrategy'
   ],
-  function (WindowTypes, PlaybackStrategy, LiveSupport, TransferFormats) {
+  function (WindowTypes, PlaybackStrategy) {
     'use strict';
 
     function shouldFailover (remainingUrls, duration, currentTime, liveSupport, windowType, transferFormat) {
       var aboutToEnd = duration && currentTime > duration - 5;
       var shouldStaticFailover = windowType === WindowTypes.STATIC && !aboutToEnd;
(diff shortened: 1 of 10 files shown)


script/models/playbackstrategy.js
```diff
@@ -0,0 +1,12 @@
+define(
+    'bigscreenplayer/models/playbackstrategy', [],
```
jlks: I'd argue that this shouldn't exist. The abstraction leaks throughout the codebase. I know a similar thing is done in hybridstrategy, but that also shouldn't exist as a module to be honest.

script/mediaresilience.js
```diff
@@ -1,16 +1,15 @@
...
   ],
-  function (WindowTypes, LiveSupport, TransferFormats) {
+  function (WindowTypes, PlaybackStrategy) {
     'use strict';
 
     function shouldFailover (remainingUrls, duration, currentTime, liveSupport, windowType, transferFormat) {
       var aboutToEnd = duration && currentTime > duration - 5;
       var shouldStaticFailover = windowType === WindowTypes.STATIC && !aboutToEnd;
-      var shouldLiveFailover = windowType !== WindowTypes.STATIC && (transferFormat === TransferFormats.DASH || liveSupport !== LiveSupport.RESTARTABLE);
+      var shouldLiveFailover = windowType !== WindowTypes.STATIC && window.bigscreenPlayer.playbackStrategy !== PlaybackStrategy.TAL && !window.bigscreenPlayer.disableLiveFailover;
```
jlks: How come we're introducing yet another configuration variable? `disableLiveFailover`

script/models/playbackstrategy.js
magnuswoody: With this piece of work, we will now need to determine whether we are in `talstrategy` or not for the failover logic. This means we will need to use these strings in three files: `mediaresilience`, `strategypicker` & `hybridstrategy`. Do you think we should keep using the plain strings?

PR #87, opened 2019-09-02
Generalised PR template

magnuswoody:
Reformatted the PR template.


tsadler1988:
I don’t think it matters that much having an internal reference visible to the public, but I’m happy to be outvoted. It might make code archaeology a bit harder though, unless the PR has all the information the ticket has.

If we do remove the Jira reference, we should also get in the habit of adding PR links in Jira.

jlks:
PR remained the same

PR #90, opened 2019-09-02
MSE autoresume at the start of SLIDING windows

magnuswoody:
📺 What

This work implements the same behaviour as TAL/Native strategies where if the player is paused at the start of the window (when viewing `WindowTypes.SLIDING` content), the player will automatically resume playback within a given cushion from the start of the window. 

Consumers of `bigscreen-player` then can begin to assume this same behaviour, regardless of strategy.

> Tickets: IPLAYERTVV1-9419

🛠 How

Implementation wise, the same utility functionality from TAL/Native strategies are used where `DynamicWindowUtils` is pulled in to allow access to the `autoResumeAtStartOfRange` functionality.


script/playbackstrategy/msestrategy.js
```diff
@@ -390,8 +394,12 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
           return isEnded;
         },
         isPaused: isPaused,
-        pause: function () {
+        pause: function (opts) {
           mediaPlayer.pause();
+          opts = opts || {};
+          if (opts.disableAutoResume !== true && windowType === WindowTypes.SLIDING) {
```
Luke-Chatburn: You have added the opts.disableAutoResume logic. Is this even used? if so, probably best you write a unit test for this bit of logic.

magnuswoody: I left it in to keep consistency between the multiple playback strategy pause methods, but can be removed if necessary.

Luke-Chatburn:
Looks good

script/dynamicwindowutils.js
```diff
@@ -59,7 +60,7 @@ define(
       addEventCallback(undefined, detectIfUnpaused);
 
       function detectIfUnpaused (event) {
-        if (event.state !== MediaPlayerBase.STATE.PAUSED) {
+        if (event.state !== MediaPlayerBase.STATE.PAUSED && event !== MediaState.PAUSED) {
```
tsadler1988: This change is a bit of a code smell, as it shows that this is working with 2 different event models. Can't think of a nice way to solve this off the top of my head I'm afraid.

tsadler1988:
Looks good thanks guys, although I did spot 1 issue (raised under #95).

PR #91, opened 2019-09-04
Spike abstraction of unpaused check.

tsadler1988:
📺 What

Make `dynamicwindowutils` agnostic of playback strategies.

🛠 How

Inject an `unpausedEventCheck` function into `autoResumeAtStartOfRange`.


jlks:
Closing as it looks to me like this was included in #90 👍

PR #93, opened 2019-09-06
Add duration to the event

mikeybyker:
📺 What

Adds the duration to the event object returned when listening for time updates.

🛠 How

Add a duration property to the object passed to the callbacks.


jlks:
Closed for now. Discussed offline months ago.

PR #99, opened 2019-10-03
Update comment

jmalk:
📺 What

I'm assuming the ended event means media has ended, not that some seeking is happening, so I've updated the comment to say that.


tsadler1988:
Forgot to say, thank you for this @jmalk.

PR #100, opened 2019-10-08
Growing window manifest refresh

matt-stephenson:
📺 What

Prevent seeking beyond the end of a growing window stream, which will result in multiple 404's and eventually a fatal error.

> Tickets: IPLAYERTVV1-9617

🛠 How

Refreshes the manifest when seeking forwards in a GROWING WINDOW asset. If the asset has ended, and the manifest contains a `mediaPresentationDuration`, the seek time will be clamped to this.

### DIFF
--- a/script-test/playbackstrategies/growingwindowrefreshertest.js
+++ b/script-test/playbackstrategies/growingwindowrefreshertest.js
@@ -5,9 +5,9 @@ require(
     function (GrowingWindowRefresher) {
       describe('GrowingWindowRefresher', function () {
         var mediaPlayerSpy;
-        var growingWindowRefresher;
         var eventHandlers;
         var dashEventCallback;
+        var seekCallbacksSpy;
 
         var dashjsMediaPlayerEvents = {
           ERROR: 'error',
@@ -25,7 +25,8 @@ require(
             eventHandlers[eventType](event);
           };
 
-          growingWindowRefresher = GrowingWindowRefresher(mediaPlayerSpy);
+          seekCallbacksSpy = jasmine.createSpy('seekCallbacksSpy');
+          GrowingWindowRefresher(mediaPlayerSpy, seekCallbacksSpy);
           jasmine.clock().install();
         });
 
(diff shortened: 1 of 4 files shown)


matt-stephenson:
Commit 578c6f9 will mean we can seek to the correct time when we fail to get the manifest on refresh. 

However, it will mean that we're NOT prevented from seeking past the end of the stream in this case.

I'm in two minds, it depends what we think is worse,
a) potential broken seek behaviour for all webcast seeking if failing to get manifest
b) rare case of seeking past the end of the stream and also failing to get manifest, and so getting 404s and a fatal error
c) this bit of code!

script/playbackstrategy/msestrategy.js
```diff
@@ -445,6 +479,8 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
           var seekToTime = getClampedTime(time, getSeekableRange());
           if (windowType === WindowTypes.SLIDING) {
             mediaElement.currentTime = (seekToTime + timeCorrection);
+          } else if (windowType === WindowTypes.GROWING && seekToTime > getCurrentTime()) {
+            refreshManifestBeforeSeek(seekToTime);
```
magnuswoody: Maybe naming should be more explicit? This also carries out the seek after the manifest has been loaded maybe `refreshManifestAndSeek(...)`?

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -849,12 +889,61 @@ require(
...
+
+        it('should publish an error event on manifest download error but there are no more sources to CDN failover to', function () {
+          var mockEvent = {
+            error: 'download',
+            event: {
+              id: 'manifest'
+            }
+          };
+
+          var noop = function () {};
```
magnuswoody: nice

script-test/playbackstrategies/growingwindowrefreshertest.js
```diff
@@ -0,0 +1,54 @@
...
+        });
+
+        it('removes MANIFEST_LOADED listener following manifest load event', function () {
+          dashEventCallback(dashjsMediaPlayerEvents.MANIFEST_LOADED, {data: {mediaPresentationDuration: 100}});
+
+          expect(mediaPlayerSpy.off).toHaveBeenCalledWith('manifestLoaded', eventHandlers.manifestLoaded);
+        });
+
+        it('should instruct the media player to refresh the manifest', function () {
+          expect(mediaPlayerSpy.refreshManifest).toHaveBeenCalledTimes(1);
```
magnuswoody: It feels like the `GrowingWindowRefresher` should be called in each test for readability

script/playbackstrategy/msestrategy.js
```diff
@@ -336,6 +368,17 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         return (mediaElement) ? mediaElement.currentTime - timeCorrection : 0;
       }
 
+      function refreshManifestBeforeSeek (seekToTime) {
+        refreshFailoverTime = seekToTime;
+        GrowingWindowRefresher(mediaPlayer, function (mediaPresentationDuration) {
+          if (!isNaN(mediaPresentationDuration)) {
+            mediaPlayer.seek(getClampedTime(seekToTime, {start: getSeekableRange().start, end: mediaPresentationDuration}));
```
magnuswoody: Worth adding any `debug.info` at this point?

script/playbackstrategy/msestrategy.js
matt-stephenson: I think I'm happy with the original!

script/playbackstrategy/msestrategy.js
matt-stephenson: Good shout. Done.

PR #114, opened 2020-01-07
Add Framework Version

jlks:
📺 What

This exposes the current framework version on the public api as `BigscreenPlayer.version`

🛠 How

This updates the framework version available at runtime based upon the package lock.

`npm run pre-release:patch|minor|major` does `npm version patch|minor|major` then updates the version JS module.


matt-stephenson:
Would be good to add to the api documentation too.

PR #117, opened 2020-01-21
Added getFrameworkVersion() to mockBigscreenPlayer

Faiza-Mazhar:
📺 What

Update mockBigscreenPlayer.
> Tickets: IPLAYERTVV1-10142

🛠 How
Added `getFrameworkVersion` to `mockBigscreenPlayer` to return mocked value of `bigscreenPlayer` version.

### DIFF
--- a/script/mockbigscreenplayer.js
+++ b/script/mockbigscreenplayer.js
@@ -6,9 +6,10 @@ define('bigscreenplayer/mockbigscreenplayer',
     'bigscreenplayer/utils/playbackutils',
     'bigscreenplayer/plugins',
     'bigscreenplayer/plugindata',
-    'bigscreenplayer/pluginenums'
+    'bigscreenplayer/pluginenums',
+    'bigscreenplayer/version'
   ],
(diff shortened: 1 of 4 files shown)


script/version.js
```diff
@@ -1,5 +1,5 @@
 define('bigscreenplayer/version',
   function () {
-    return '3.8.0';
```
jlks: Looks like you haven't correctly run ``npm run pre-publish:patch` here, otherwise the `package.json` and package-lock.json` would also be updated.

script/mockbigscreenplayer.js
```diff
@@ -264,6 +267,9 @@ define('bigscreenplayer/mockbigscreenplayer',
       getPlayerElement: function () {
         return;
       },
+      getFrameworkVersion: function () {
+        return version;
```
jlks: You can probably not bother mocking this out and introduce Version as a dependency for direct usage like in bigscreenplayer.js

Faiza-Mazhar: Thank you. I changed it.

PR #118, opened 2020-01-21
Improve the development experience

jlks:
📺 What

Currently standalone development has to go through the example app, which is clunky to update on the fly and requires a rebuild each time. This just leverages webpack-dev-server to do the hot reloading for us.

🛠 How

Use webpack-dev-server. Enable quicker rapid prototyping locally without an application integration.

`npm start` will get you a webpack-dev-server with hot module replacement

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,20 +13,14 @@
       app: './development/index.js'
     },
     plugins: [
-      // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
       new CleanWebpackPlugin(),
       new HtmlWebpackPlugin({
-        // Required
         inject: false,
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -0,0 +1,42 @@
+
+  const path = require('path');
+  const HtmlWebpackPlugin = require('html-webpack-plugin');
+  const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+
+  module.exports = {
+   mode: 'development',
```
Luke-Chatburn: Can we add a `watch: true` here as well?

jlks: That is enabled by [default ](https://webpack.js.org/configuration/watch/#watch)

PR #119, opened 2020-01-22
Correct bitrate plugin metric

jlks:
📺 What

This emits the full playbackBitrate from `msestrategy`

🛠 How

The current representation bitrates for video and audio are combined and emitted as `playbackBitrate` through Plugins.interface.onPlayerInfoUpdated.

### DIFF
--- a/script-test/playbackstrategies/msestrategytest.js
+++ b/script-test/playbackstrategies/msestrategytest.js
@@ -799,7 +799,7 @@ require(
           mockPluginsInterface.onPlayerInfoUpdated.calls.reset();
         });
 
-        it('should call plugins with video playback bitrate', function () {
+        it('should call plugins with the combined playback bitrate', function () {
           setUpMSE();
           mockDashInstance.getBitrateInfoListFor.and.returnValue([{ bitrate: 1024000 }, { bitrate: 200000 }, { bitrate: 3000000 }]);
           mseStrategy.load(null, 0);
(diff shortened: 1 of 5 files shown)


script/playbackstrategy/msestrategy.js
```diff
@@ -160,25 +160,51 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         DebugTool.info('Manifest validity changed. Duration is: ' + event.newDuration);
       }
 
+      function onStreamInitialised () {
+        emitPlayerInfo();
+      }
+
+      function emitPlayerInfo () {
+        if (mediaKind === MediaKinds.VIDEO) {
+          playerMetadata.playbackBitrate = currentPlaybackBitrate('video') + currentPlaybackBitrate('audio');
```
matt-stephenson: Could we use the `MediaKinds` enums here

script/playbackstrategy/msestrategy.js
```diff
@@ -160,25 +160,51 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         DebugTool.info('Manifest validity changed. Duration is: ' + event.newDuration);
       }
 
+      function onStreamInitialised () {
+        emitPlayerInfo();
```
matt-stephenson: Could we just have this function as the handler directly on the event? Happy if you prefer this as more readable though.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -792,13 +801,13 @@ require(
 
         it('should call plugins with video playback bitrate', function () {
```
matt-stephenson: Maybe rename this test to make it clear this is the video and audio bitrate case?

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -812,7 +821,7 @@ require(
           };
 
           setUpMSE();
-          mockDashInstance.getBitrateInfoListFor.and.returnValue([{ bitrate: 1000 }, { bitrate: 2048 }, { bitrate: 3000 }]);
+          mockDashInstance.getBitrateInfoListFor.and.returnValue([{ bitrate: 1024000 }, { bitrate: 2000000 }, { bitrate: 3000000 }]);
```
matt-stephenson: Is this this test case still true? Possibly misunderstanding, but if the audio bitrate changes within a video, would we not want to call the plugin, given the client is reporting combined bitrate now?

jlks: This is still valid, I changed the mock values to represent something more reasonable. The event is called once, mediaKinds is video and both the video and audio bitrates are queried and combined. Rather than make more work for myself with the tests I just changed what was required.

script/playbackstrategy/msestrategy.js
jlks: Nope, it has another usage.

script-test/playbackstrategies/msestrategytest.js
matt-stephenson: Ah, I'm still not understanding. The name of the test is:
"_should not call plugins with audio playback bitrate when mediaKind is video_"
But now we do care when the audio playback bitrate changes right? So why would we not call the plugins with this updated information?

PR #127, opened 2020-02-20
Avstats bug seek event

Faiza-Mazhar:
📺 What
For most MSE devices, we throw TWO buffering state changes from bigscreen-player, one on the media element's 'waiting' event, and another on the media element's 'seeking' event. The first one will be suppressed when seeking, but the second will not. We need to throw one `waiting` event to stop reporting false buffering. 

> [Tickets: IPLAYERTVV1-10185](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-10185)

🛠 How
Set seeking state when `setCurrentTime` is called in `msestrategy`.  Throw only the first seeking/waiting event when we are in the seeking state.

### DIFF
--- a/script/playbackstrategy/msestrategy.js
+++ b/script/playbackstrategy/msestrategy.js
@@ -33,6 +33,9 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       var mediaMetrics;
       var dashMetrics;
 
+      var publishedSeekEvent = false;
+      var isSeeking = false;
+
       var playerMetadata = {
         playbackBitrate: undefined,
         bufferLength: undefined,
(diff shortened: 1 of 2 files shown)


script/playbackstrategy/msestrategy.js
```diff
@@ -68,10 +71,16 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
 
       function onBuffering () {
         isEnded = false;
-        publishMediaState(MediaState.WAITING);
+        if (isSeeking && !publishedSeekEvent) {
+          publishMediaState(MediaState.WAITING);
+          publishedSeekEvent = true;
+        } else if (!isSeeking) {
+          publishMediaState(MediaState.WAITING);
+        }
```
tsadler1988: This would work too, but is arguably less readable.
```
if (!isSeeking || !publishedSeekEvent) {
  publishMediaState(MediaState.WAITING);
  publishedSeekEvent = true;
}
```

PR #128, opened 2020-02-21
Multiple seek event bug

Faiza-Mazhar:
📺 What
For most MSE devices, we throw TWO buffering state changes from `bigscreen-player`, one on the media element's `waiting` event, and another on the media element's `seeking` event. The first one will be suppressed when seeking, but the second will not. We need to throw one `WAITING` event to stop reporting false buffering.

[Tickets: IPLAYERTVV1-10185](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-10185)

🛠 How
Set seeking state when `setCurrentTime` is called in `msestrategy`. Throw onw `WAITING` for the first seeking/waiting event when we are in the seeking state.

### DIFF
--- a/script/playbackstrategy/msestrategy.js
+++ b/script/playbackstrategy/msestrategy.js
@@ -33,6 +33,9 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
       var mediaMetrics;
       var dashMetrics;
 
+      var publishedSeekEvent = false;
+      var isSeeking = false;
+
       var playerMetadata = {
         playbackBitrate: undefined,
         bufferLength: undefined,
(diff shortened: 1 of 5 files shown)


script/playbackstrategy/msestrategy.js
```diff
@@ -68,10 +71,24 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
 
       function onBuffering () {
         isEnded = false;
-        publishMediaState(MediaState.WAITING);
+        if (!isSeeking || !publishedSeekEvent) {
```
tsadler1988: Assuming you will clean up this debugging before merging? If so, should this be a draft PR?

Faiza-Mazhar: Yes, It's only added for testing purposes. 

Though we can't change a PR to Draft PR.

PR #130, opened 2020-03-26
Add a guard to eventCallbacks

jlks:
📺 What
Spotted in automated testing telemetry. Calling tearDown will set `eventCallbacks` to `undefined` causing a type error when a media element emits an event.

Spotted during
> Tickets: IPLAYERTVV1-9375

🛠 How

Guard iterating through the `eventCallbacks` if the array is `undefined`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bigscreen-player",
-  "version": "3.13.0",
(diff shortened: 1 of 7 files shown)


tsadler1988:
I believe `cehtml` and `msestrategy` have the same problem.

PR #131, opened 2020-04-07
Return `undefined` from `getCurrentTime()` when it is unknown

tjenkinson:
📺 What

Currently `getCurrentTime()` has a default value of 0 in several cases. This is not ideal because it makes it impossible to know if it truly is time 0, or if the time is not known yet.

It also sometimes returns the wrong value.

An example where this is an issue is if you start a live stream from the live point. Currently you will get back the incorrect time whilst the stream loads, and then it will jump to the correct value.

With this PR when the strategy is still loading (hasn't changed the state to something other than `WAITING`) `getCurrentTime()` will now always return

- the value of the latest `setCurrentTime` request or
- `initialPlaybackTime` or
- `undefined`

I chose to put this logic in 'bigscreenplayer.js' instead of the individual strategies because I think this is something that benefits them all.

This also means a call to `getCurrentTime()` in the same stack as `setCurrentTime()` is guaranteed to be what you expect even if the update happens asynchronously in the strategy.

This PR also contains a fix in the `msestrategy` `getCurrentTime()` to prevent it returning incorrect values when the media element isn't ready. This change isn't strictly needed given the other change in 'bigscreenplayer.js' prevents it causing an issue, but it's probably good to fix anyway.

> N/A (but can create one)

🛠 How

- Introduce a `SeekState` enum which can be `NONE`, `QUEUED` or `IN_FLIGHT`. When a seek is `QUEUED` or `IN_FLIGHT` we now return the time we're seeking to from `getCurrentTime()` instead of calling through to the strategy.
- In `msestrategy` return `undefined` instead of `0` from `getCurrentTime` when the time is unknown.
- In `msestrategy` return `undefined` when the media element [`readyState`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState) is at least `HAVE_METADATA`. Before this dashjs has not set the current time to near the live point meaning it would return the wrong value.
- Update/add tests

 ✅ Acceptance criteria
- [ ] `getCurrentTime()` returns `undefined` whilst a stream is loading with no `initialPlaybackTime` using the mse strategy
- [ ] `getCurrentTime()` always returns the time it was set to immediately after `setCurrentTime()`
- [ ] `getCurrentTime()` initially returns the value of `initialPlaybackTime`.
- [ ] `getCurrentTime()` returns the expected time when the media state is not WAITING

* [ ] [Mark correctly if it is covered in this PR or not]

### DIFF
--- a/script-test/playbackstrategies/msestrategytest.js
+++ b/script-test/playbackstrategies/msestrategytest.js
@@ -26,8 +26,8 @@ require(
     var mockPlugins;
     var mockPluginsInterface;
     var mockDynamicWindowUtils;
-    var mockAudioElement = document.createElement('audio');
-    var mockVideoElement = document.createElement('video');
+    var mockAudioElement = document.createElement('div');
+    var mockVideoElement = document.createElement('div');
     var mockRefresher;
     var testManifestObject;
     var timeUtilsMock;
@@ -512,16 +512,17 @@ require(
         it('returns the correct time from the DASH Mediaplayer', function () {
           setUpMSE();
           mockVideoElement.currentTime = 10;
+          mockVideoElement.readyState = 1;
 
           mseStrategy.load(null, 0);
 
           expect(mseStrategy.getCurrentTime()).toBe(10);
         });
 
-        it('returns 0 when MediaPlayer is undefined', function () {
+        it('returns undefined when MediaPlayer is undefined', function () {
           setUpMSE();
 
-          expect(mseStrategy.getCurrentTime()).toBe(0);
+          expect(mseStrategy.getCurrentTime()).toBe(undefined);
         });
       });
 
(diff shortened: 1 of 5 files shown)


script-test/playbackstrategies/msestrategytest.js
```diff
@@ -26,8 +26,8 @@ require(
     var mockPlugins;
     var mockPluginsInterface;
     var mockDynamicWindowUtils;
-    var mockAudioElement = document.createElement('audio');
-    var mockVideoElement = document.createElement('video');
+    var mockAudioElement = document.createElement('div');
```
tjenkinson, line 29: Setting `mockVideoElement.readyState` does not work if it's a real element

PR #135, opened 2020-05-20
Auto Resume Live Asset for MSE Devices 

jamierbower:
📺 What

Seeking to the beginning of the sliding window on a MSE device will resume play from a paused state.

> Tickets: [IPLAYERTVV1-9679](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-9679)

🛠 How

When seeking to a point in the video, we check if the new time is at the start of the sliding window. If it is at the start and we are in a paused state then play the asset.

[MANDATORY for cross-team reviews]
* [X] [The asset resumes when we seek to the beginning]

### DIFF
--- a/script/playbackstrategy/legacyplayeradapter.js
+++ b/script/playbackstrategy/legacyplayeradapter.js
@@ -247,6 +247,7 @@ define('bigscreenplayer/playbackstrategy/legacyplayeradapter',
           mediaPlayer.initialiseMedia('video', mediaSources.currentSource(), mimeType, playbackElement, setSourceOpts);
           if (mediaPlayer.beginPlaybackFrom && !isPlaybackFromLivePoint) {
             currentTime = startTime;
+            DebugTool.keyValue({key: 'initial-playback-time', value: startTime + timeCorrection});
             mediaPlayer.beginPlaybackFrom(startTime + timeCorrection || 0);
           } else {
             mediaPlayer.beginPlayback();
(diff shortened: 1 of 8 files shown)


script/playbackstrategy/legacyplayeradapter.js
```diff
@@ -247,6 +247,7 @@ define('bigscreenplayer/playbackstrategy/legacyplayeradapter',
           mediaPlayer.initialiseMedia('video', mediaSources.currentSource(), mimeType, playbackElement, setSourceOpts);
           if (mediaPlayer.beginPlaybackFrom && !isPlaybackFromLivePoint) {
             currentTime = startTime;
+            DebugTool.keyValue({key: 'initial-playback-time', value: startTime + timeCorrection});
```
Luke-Chatburn, line 250: Is it worth adding the || 0 here to keep the log consistent with the code.

script/playbackstrategy/msestrategy.js
```diff
@@ -373,6 +373,7 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         }
 
         if (windowType === WindowTypes.SLIDING) {
+          DebugTool.keyValue({key: 'initial-playback-time', value: startTime});
```
Luke-Chatburn, line 376: Again, probably best to keep the log the same.

jlks: it is.

script/playbackstrategy/legacyplayeradapter.js
jlks: I want to see when the value is out of range or undefined.

script/playbackstrategy/msestrategy.js
Luke-Chatburn: The time passed into dashjs is parsed, the log isn't. i know it will be the same value just with decimal places. As there is a reason we parse the value.

jlks: I get you now!

PR #137, opened 2020-06-03
Add ability to notified to subtitles being turned on/off

tjenkinson:
📺 What

This adds `registerForSubtitleChanges()`. The listener is called whenever `setSubtitlesEnabled` is.

> Tickets: IPLAYERTVV1-11166

🛠 How
Implement `registerForSubtitleChanges` and `unregisterForSubtitleChanges` using the other similar functions as an example.

 ✅ Acceptance criteria

* [x] The callback is called when subtitles are turned on/off and no longer called when it's unregistered.

### DIFF
--- a/script/bigscreenplayer.js
+++ b/script/bigscreenplayer.js
@@ -18,6 +18,7 @@ define('bigscreenplayer/bigscreenplayer',
     function BigscreenPlayer () {
       var stateChangeCallbacks = [];
       var timeUpdateCallbacks = [];
+      var subtitleCallbacks = [];
 
       var mediaKind;
       var initialPlaybackTimeEpoch;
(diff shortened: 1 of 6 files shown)


script/bigscreenplayer.js
```diff
@@ -253,6 +265,9 @@ define('bigscreenplayer/bigscreenplayer',
         },
         setSubtitlesEnabled: function (value) {
           playerComponent.setSubtitlesEnabled(value);
+          subtitleCallbacks.forEach(function (callback) {
```
jlks, line 268: Something that just came to mind. You might consider implementing this in `playerComponent` due to the `enableSubtitles` flag that is passed into `playerComponent` when it is first used. I don't think this would get the initial state emitted.

tjenkinson: good catch. I pushed a commit for this. It keeps it in this file for now because this is where the other event like callbacks are  but can move it into the player component if you want.

PR #144, opened 2020-06-16
Auto resume after seek whilst paused

Luke-Chatburn:
📺 What

This fixes auto resume not triggering for MSE and native seekable whilst paused and seeking to anywhere other than the start.

> Tickets: IPLAYERTVV1-11210

🛠 How

- Starts auto resume timer if paused and seeking
(MSE is triggered after a 'seeked' event)
(Native seekable is on playFrom if postseekstate was paused.)
- Stops STATUS events clearing auto resume timer for native implementations
- Verifies that only SLIDING windows start the timeout

### DIFF
--- a/script-test/playbackstrategies/modifiers/live/seekabletest.js
+++ b/script-test/playbackstrategies/modifiers/live/seekabletest.js
@@ -351,6 +351,8 @@ require(
 
           seekableMediaPlayer.playFrom(50);
 
+          seekableMediaPlayer.pause();
+
           jasmine.clock().tick(42 * 1000);
 
           expect(player.resume).toHaveBeenCalledTimes(1);
--- a/script-test/playbackstrategies/modifiers/live/seekabletest.js
+++ b/script-test/playbackstrategies/modifiers/live/seekabletest.js
@@ -1,9 +1,10 @@
 require(
   [
     'bigscreenplayer/playbackstrategy/modifiers/mediaplayerbase',
-    'bigscreenplayer/playbackstrategy/modifiers/live/seekable'
+    'bigscreenplayer/playbackstrategy/modifiers/live/seekable',
+    'bigscreenplayer/models/windowtypes'
   ],
-    function (MediaPlayerBase, SeekableMediaPlayer) {
+    function (MediaPlayerBase, SeekableMediaPlayer, WindowTypes) {
       var sourceContainer = document.createElement('div');
       var player;
       var seekableMediaPlayer;
@@ -20,8 +21,8 @@ require(
         }
       }
 
-      function initialiseSeekableMediaPlayer (config) {
-        seekableMediaPlayer = SeekableMediaPlayer(player, config);
+      function initialiseSeekableMediaPlayer (config, windowType) {
+        seekableMediaPlayer = SeekableMediaPlayer(player, config, windowType);
       }
 
       describe('Seekable HMTL5 Live Player', function () {
@@ -172,7 +173,7 @@ require(
             jasmine.clock().install();
             jasmine.clock().mockDate();
 
-            initialiseSeekableMediaPlayer();
+            initialiseSeekableMediaPlayer(undefined, WindowTypes.SLIDING);
 
             player.getSeekableRange.and.returnValue({start: 0});
             player.getCurrentTime.and.returnValue(20);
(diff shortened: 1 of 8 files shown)


script-test/playbackstrategies/modifiers/live/seekabletest.js
```diff
@@ -326,6 +339,22 @@ require(
...
           expect(player.resume).not.toHaveBeenCalledTimes(1);
         });
+
+        it('Should auto resume when paused after a seek', function () {
+          player.getSeekableRange.and.returnValue({start: 0});
+          player.getCurrentTime.and.returnValue(100);
+
+          startPlaybackAndPause(100, false);
+
+          player.getCurrentTime.and.returnValue(50);
```
tsadler1988, line 349: Surely this should happen after `playFrom`?

script-test/playbackstrategies/modifiers/live/seekabletest.js
```diff
@@ -326,6 +339,22 @@ require(
...
+          player.getState.and.returnValue(MediaPlayerBase.STATE.PAUSED);
```
tsadler1988, line 350: Shouldn't this be a call to `pause` after calling `playFrom`  rather than forcing the state?

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -765,25 +763,20 @@ require(
             expect(timeUtilsMock.calculateSlidingWindowSeekOffset).toHaveBeenCalledTimes(1);
           });
 
-          it('should autoresume when paused and seeking to the start of the sliding window seekable range', function () {
-            mockDynamicWindowUtils.shouldAutoResume.and.returnValue(true);
+          it('should start auto resume timeout when paused and seeking', function () {
             mockDashInstance.isPaused.and.returnValue(true);
 
-            mseStrategy.pause();
-            mseStrategy.setCurrentTime(0);
```
tsadler1988, line 773: Feel like we should keep these 2 lines in for it to be representative. `seeked` wouldn't get fired for no reason like it appears to be here.

script-test/playbackstrategies/msestrategytest.js
```diff
@@ -765,25 +763,20 @@ require(
...
-            expect(mockDashInstance.play).toHaveBeenCalledTimes(1);
+            expect(mockDynamicWindowUtils.autoResumeAtStartOfRange).toHaveBeenCalledTimes(1);
           });
 
-          it('should not try to autoresume when playing and seeking to the start of the sliding window seekable range', function () {
-            mockDynamicWindowUtils.shouldAutoResume.and.returnValue(true);
+          it('should not try to autoresume when playing and seeking', function () {
             mockDashInstance.isPaused.and.returnValue(false);
 
-            mseStrategy.setCurrentTime(0);
```
tsadler1988, line 783: Similar to my above comment, I don't think this is representative.

script/playbackstrategy/modifiers/live/restartable.js
```diff
@@ -50,14 +50,10 @@ define(
...
 
-      function resume () {
-        mediaPlayer.resume();
-      }
-
       function pause (opts) {
         mediaPlayer.pause();
         opts = opts || {};
-        if (opts.disableAutoResume !== true) {
+        if (opts.disableAutoResume !== true && windowType === WindowTypes.SLIDING) {
```
tsadler1988, line 56: No tests for this change?

script-test/playbackstrategies/modifiers/live/seekabletest.js
Luke-Chatburn: It can't because playFrom kicks off dynamicWindowUtils which is checking getCurrentTime

script-test/playbackstrategies/modifiers/live/seekabletest.js
Luke-Chatburn: HTML5 Player is being spied on, pause isn't triggering state changes.

script/playbackstrategy/modifiers/live/restartable.js
Luke-Chatburn: It was sliding by default. I have added one to test if it isn't a sliding window it doesn't get called.

script-test/playbackstrategies/modifiers/live/seekabletest.js
tsadler1988: As discussed, this sort of change would be difficult given the state they're in, so we'll be pragmatic rather than perfectionist.

script-test/playbackstrategies/modifiers/live/seekabletest.js
tsadler1988: As discussed, this sort of change would be difficult given the state the tests are in, so we'll be pragmatic rather than perfectionist.

PR #148, opened 2020-06-22
Remove auto resume function

jlks:
📺 What

This removes the `shouldAutoResume` function that was missed in the tidy up of #144 .

🛠 How

Remove the function and accompanying tests.


jlks:
This needs the mock removing from msestrategytest too

PR #149, opened 2020-06-26
Proposal for contribution guidelines

tsadler1988:
📺 What

Add contribution guidelines

> Tickets: N/A

🛠 How

CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,36 @@
+# Contributing
+
+We welcome and encourage contributions. We are open minded to suggestions, encourage debate, and aim to build a consensus; if there is no consensus, maintainers will have the final say.
+
+Unless an Issue or Pull Request can be resolved within 2 working days, we will raise an internal ticket for it on our backlog and note this on GitHub.
+
+## Opening an Issue
+
+For any queries, suggestions, or bug reports, please raise an issue.
+
+## Opening a Pull Request
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
+# Contributing
+
+We welcome and encourage contributions. We are open minded to suggestions, encourage debate, and aim to build a consensus; if there is no consensus, maintainers will have the final say.
```
matt-stephenson, line 3: Worth saying who the 'we' is in this first usage?

matt-stephenson, line 3: I read the 'maintainers will have the final say' and felt it was a bit 'strongly worded' for the opening sentence. Could we:
a) make it less so
b) just not write this, as it's essentially covered below in the review/test/release process (maybe make that section more explicit to explain the point)

tsadler1988: I'd have thought that was fairly clear. Not sure what else we'd put:
- BBC is too generic
- Honey Badger only makes sense internally
- Bigscreen Player maintainers is a bit wordy

tsadler1988: Yes I wasn't sure about putting this in. Perhaps aimed more at internal contributors - I think it's fairly accepted that maintainers have the final say in Open Source. Best to just remove it?

matt-stephenson: Yep, fair point 👍

matt-stephenson: 👍

PR #160, opened 2020-07-24
Build output improvements

jlks:
📺 What

Tidy up build distribution output for NPM.
> Tickets: N\A

🛠 How

- build output to the dist folder
- TODO: only publish from this folder


jlks:
`pre-release` scripts needs version bumping removing before progressing.

PR #170, opened 2020-09-02
Make subtitles testable

jlks:
📺 What

This decomposes `captions.js` into its component parts so that they are testable in isolation.

> Tickets: [IPLAYERTVV1-11504](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-11504)

🛠 How

Add a number of test files around:
- Transforming xml subtitles into our internal representation.
- Creating individual subtitles to be added to the DOM.

Decompose the functionality from `captions.js` in to the separate modules.


ShiningTrapez (approved):
LGTM

development/index.js
```diff
@@ -27,6 +27,18 @@ require (['bigscreenplayer/bigscreenplayer'], function(BigscreenPlayer){
...
+  function setUpCaptionsContainerCSS() {
+    var captionsContainer = document.getElementById('playerCaptionsContainer');
+    captionsContainer.style.position = 'absolute';
+    captionsContainer.style.top = '80%';
+    captionsContainer.style.right = '50%';
+  }
+  
+  function normaliseVolume () {
+    var videoElement = document.getElementsByTagName('video')[0];
+    videoElement.volume = 0.1;
```
Luke-Chatburn, line 39: What is this change for? When we tested this in rDot we never saw a value less than 1. Also, why are we setting it to the lowest possible value? This would make us inconsistent with other players and normal linear tv. 'Playing normal TV with acceptable volume and no problem, go to iplayer and it is too low, so turn it up. Go back to normal TV and it is deafeningly loud!'

Luke-Chatburn (changes requested):

development/index.js
jlks: Well it is the dev environment script. The test asset I was using was rather loud when trying to remote pair with headphones.

Luke-Chatburn: Didn't notice the dev script :D

Luke-Chatburn (approved):

PR #172, opened 2020-10-05
Resizer

matt-stephenson:
📺 What

Adds ability to resize the playback element.

> Tickets: IPLAYERTVV1-11651

🛠 How

Modify the style on the playback element with a top, bottom, width, height, zIndex.
Allowed sizes are not sanity checked.
Subtitles are automatically disabled upon resize, and re-enabled (if originally enabled) when clearing the resize.

✅ Testing

| ---- | ---- |

Manual testing in development environment.


script/bigscreenplayer.js
```diff
@@ -142,13 +145,24 @@ define('bigscreenplayer/bigscreenplayer',
...
       return {
-        init: function (playbackElement, bigscreenPlayerData, newWindowType, enableSubtitles, callbacks) {
+        init: function (newPlaybackElement, bigscreenPlayerData, newWindowType, enableSubtitles, callbacks) {
           Chronicle.init();
+          resizer = Resizer();
           DebugTool.setRootElement(playbackElement);
           DebugTool.keyValue({key: 'framework-version', value: Version});
           windowType = newWindowType;
           serverDate = bigscreenPlayerData.serverDate;
+          playbackElement = newPlaybackElement;
```
Luke-Chatburn, line 165: Instead of storing a reference to element can we not get it from the DOM when we require it?

script/bigscreenplayer.js
```diff
@@ -270,13 +286,19 @@ define('bigscreenplayer/bigscreenplayer',
...
-          callSubtitlesCallbacks(value);
+        resize: function (top, left, width, height, zIndex) {
+          subtitlesHidden = isSubtitlesEnabled();
+          setSubtitlesEnabled(false);
+          resizer.resize(playbackElement, top, left, width, height, zIndex);
         },
-        isSubtitlesEnabled: function () {
-          return playerComponent ? playerComponent.isSubtitlesEnabled() : false;
+        clearResize: function () {
+          if (subtitlesHidden) {
```
Luke-Chatburn, line 295: Can we not just use isSubtitlesEnabled() to turn them back on instead of having more state with subtitlesHidden?

matt-stephenson: No, because the `resize` function makes these false, so if we just used `isSubtitlesEnabled` when clearing the resize, this would always be false.

script/bigscreenplayer.js
matt-stephenson: Would rather store in this instance. We take the whole playback element, so we'd need to know the id of this. It's not the video element that is resized, it's the whole containing div.

Luke-Chatburn (approved):

PR #174, opened 2020-10-13
IPLAYERTVV1 10894 Call Success Callback when BSP Ready

ShiningTrapez:
📺 What

> Tickets: IPLAYERTVV1-10894

🛠 How

Add a simple latching system which responds to events and calls the success callback when we have a valid current time and/or seekable range (Depending on asset type) 

✅ Testing

See https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-10894

| ---- | ---- |

👀  See

May induce slight lag in UI first paint, likely not to be noticeable to end users.


script/readyhelper.js
```diff
@@ -0,0 +1,83 @@
...
+        return false;
+      }
+
+      function validateLiveTime (currentTime, seekableRange) {
+        if (liveSupport === LiveSupport.PLAYABLE) {
+          return currentTime >= 0;
+        }
+
+        if (isValidSeekableRange(seekableRange)) {
+          var currTimeGtStart = currentTime >= seekableRange.start;
```
Luke-Chatburn, line 58: currentTimeWithinRange feels a more suitable option since we are not interested in the variables individually.

script/readyhelper.js
```diff
@@ -0,0 +1,83 @@
+define(
+  'bigscreenplayer/readyhelper', [
+    'bigscreenplayer/models/mediastate',
+    'bigscreenplayer/models/windowtypes',
+    'bigscreenplayer/models/livesupport'
+  ],
+  function (MediaState, WindowTypes, LiveSupport) {
+    var ReadyHelper = function (initialPlaybackTime, windowType, liveSupport, callback) {
```
Luke-Chatburn, line 8: var ReadyHelper = function is the same as function ReadyHelper. We should keep our code consistent.

script/readyhelper.js
```diff
@@ -0,0 +1,83 @@
...
+      var ready = false;
+
+      var callbackWhenReady = function (evt) {
```
Luke-Chatburn, line 11: Same here for consistent function declarations.

script/bigscreenplayer.js
```diff
@@ -80,6 +83,10 @@ define('bigscreenplayer/bigscreenplayer',
         if (evt.data.duration) {
           DebugTool.keyValue({key: 'duration', value: evt.data.duration});
         }
+
+        if (playerComponent && readyHelper) {
```
Luke-Chatburn, line 87: They have to call BSP.init, so PlayerComponent and readyHelper must always be available?

Luke-Chatburn (changes requested):
Not to sure readyHelper is the correct way to describe your logic. Not got a better option in my head though :D

script/readyhelper.js
ShiningTrapez: This style is used elsewhere, which is why I used it in the first place! See for example https://github.com/bbc/bigscreen-player/blob/master/script/playercomponent.js#L17

script/bigscreenplayer.js
ShiningTrapez: If we don't do this, you get race conditions; this is for example what caused the subtitles bug on Native Devices in iPlayer, due to initialization issues

ShiningTrapez:
> Not to sure readyHelper is the correct way to describe your logic. Not got a better option in my head though :D

I was considering breaking this out into a more generic "Latch" module which would just contain the logic incase we need false until true in the future, but I thought that might be too much abstraction for this ticket

script/readyhelper.js
jamierbower: do you mean pulling the range checking logic into a separate function?

Luke-Chatburn: No, just a single variable instead of two.

ShiningTrapez: We split it up because it was really long as one variable, and went quite a bit off the side of the screen

ShiningTrapez: OK, I condensed it down, I think it's still readable enough; I think at one point the logic was a bit more complex there

Luke-Chatburn (approved):

PR #180, opened 2020-11-12
Introduce a subtitles abstraction

jlks:
📺 What

Encapsulate the state of subtitles in one place.

> Tickets: IPLAYERTVV1-11801

🛠 How

- Move the enabled and availability states to sit with the implementation in a module rather than separate variables to be reasoned about in `PlayerComponent`. 
- `CaptionsContainer` is renamed to `LegacySubtitles`.
- Extract subtitles file loading out of `renderer` into the `Subtitles` abstraction. 
- Change `LoadUrl` to be usable with a callback pattern in all cases. This no longer exposes the internal XHR as a return value.
- Improved api naming.


Luke-Chatburn (approved):

script/subtitles/subtitles.js
```diff
@@ -0,0 +1,79 @@
...
+        subtitlesEnabled = true;
+        if (available() && subtitlesContainer) {
+          subtitlesContainer.start();
+        }
+      }
+
+      function stop () {
+        subtitlesEnabled = false;
+        if (available() && subtitlesContainer) {
+          subtitlesContainer.stop();
```
jamierbower, line 45: Would it be possible to setting the enabled state and stopping and starting the subtitles in separate functions? this would allow you to have subtitles enabled, but hidden (eg in a resized window)

jamierbower (changes requested):

script/subtitles/subtitles.js
jamierbower: Agreed to do this in a later iteration

jamierbower (approved):

PR #182, opened 2020-11-17
Bug fix - only attempt to load subtitles file when there is a url to do so

matt-stephenson:
📺 What

Only attempt to load the subtitles file when there is a URL to do so.

> Tickets: IPLAYERTVV1-11801

🛠 How

Check for a truthy subtitles url before attempting to load.


ShiningTrapez (approved):
LGTM

PR #184, opened 2020-12-01
Implement subtitles using smp fork of imscJS

jlks:
📺 What

This implements subtitles in line with the [BBC subtitle guidelines](http://bbc.github.io/subtitle-guidelines/) and provides a fallback to the old implementation for devices that are not capable or performant.

> Tickets: IPLAYERTVV1-11770

🛠 How

Take imscJS library and implement alongside 'legacy' subtitles. A config value is used to determine which subtitles rendering method is used, via a dynamic require.

✅ Testing 

| ---- | ---- |

Manual testing to ensure new rendering works across a wide range of devices.


script/subtitles/imscsubtitles.js
```diff
@@ -0,0 +1,100 @@
...
+
+          previousSubtitlesIndex = subtitlesIndex;
+        }
+      }
+
+      function start () {
+        if (xml && times.length > 0) {
+          updateInterval = setInterval(function () {
+            update(mediaPlayer.getCurrentTime());
+          }, 750);
```
jamierbower, line 78: This could be a var at the top of the file, if you wanted to ever change the frequency it might not be obvious it's in the start function

jamierbower (approved):

PR #188, opened 2021-01-04
Add new playback strategy - basicStrategy

matt-stephenson:
📺 What

Adds a new playback strategy - `basicStrategy`.

> Tickets: IPLAYERTVV1-11669

🛠 How

Adds new strategy file. This strategy uses the devices native playback capabilities, but differs from the `nativeStrategy` in that it does not contain specific playback state or work-arounds for non-compliant browsers. Playback is fully event driven, and any state is determined directly from the browsers media element.

✅ Testing

| ---- | ---- |

Tested with extensive manual testing.

 ✅ Acceptance criteria
      ✅ Feature parity with existing Native Strategy


jlks:
Mind updating the wiki on configuration when this is in? https://github.com/bbc/bigscreen-player/wiki/Playback-Strategy

script/readyhelper.js
```diff
@@ -54,11 +54,11 @@ define(
           return currentTime >= 0;
         }
 
-        if (isValidSeekableRange(seekableRange)) {
-          return currentTime >= seekableRange.start &&
```
jlks, line 58: Does this fix a bug outside of this PR or should this check really be inside `isValidSeekableRange`?

matt-stephenson:
> Mind updating the wiki on configuration when this is in? https://github.com/bbc/bigscreen-player/wiki/Playback-Strategy

👍 yep, will add this to the ticket to do. Thanks.

script/readyhelper.js
matt-stephenson: The checking of currentTime being within the seekable range was breaking things on occasions for this new basicStrategy. My understanding is this (currentTime lies within the seekableRange) isn't a spec requirement, so we decided to relax that rule here (will apply to all strategies of course).

ShiningTrapez (approved):
LGTM

PR #189, opened 2021-01-12
Subtitle Customisation Defaults

jamierbower:
📺 What

Allow subtitles customisation options to passed into BSP on initialisation.

> Tickets: [IPLAYERTVV1-11933](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-11933)

🛠 How

Add a subtitles customisation object to bigscreenPlayerData.
Take `bbc/imscJS#v1.0.3` which has an API for customising subtitles appearance.

✅ Testing 

| ---- | ---- |

### DIFF
--- a/script/subtitles/imscsubtitles.js
+++ b/script/subtitles/imscsubtitles.js
@@ -33,7 +33,7 @@ define('bigscreenplayer/subtitles/imscsubtitles',
         var customStyles = {};
 
         if (opts.backgroundColour) {
-          customStyles.spanBackgroundColorAdjust = {transparent: opts.backgroundColour, black: opts.backgroundColour};
+          customStyles.spanBackgroundColorAdjust = {transparent: opts.backgroundColour};
         }
 
         if (opts.fontFamily) {
--- a/script/subtitles/imscsubtitles.js
+++ b/script/subtitles/imscsubtitles.js
@@ -26,7 +26,7 @@ define('bigscreenplayer/subtitles/imscsubtitles',
         Plugins.interface.onSubtitlesTransformError();
       }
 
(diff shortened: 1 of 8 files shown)


script/subtitles/imscsubtitles.js
```diff
@@ -25,6 +26,22 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
       }
 
+      // Opts: { backgroundColour: string (css colour, hex), fontFamily: string }
+      function transformStyleOptions (opts) {
+        if (opts === undefined) return;
+
+        var customStyles = {};
+
+        if (opts.backgroundColour) {
+          customStyles.spanBackgroundColorAdjust = {transparent: opts.backgroundColour, black: opts.backgroundColour};
```
nigelmegitt, line 36: Why also transform already black backgrounds? If you omit that the result should be the same.

jamierbower: this implementation doesn't have knowledge if the colour we are transforming to, as it's passed in. hypothetically you might want to set it as another colour

nigelmegitt: right, but then always transforming _black_ won't work since the implementation also doesn't know the colour it is transforming _from_. There are some old subs files around that do use "interesting" background colours.

nigelmegitt: This may be an argument for a customisation that can map "any", which is something I've been wondering about already.

Then with a mapping table:
```json
{
   "*": "black",
   "black": "navy"
}
```

where `*` is some wildcard that matches everything,

should a black background passed in end up black or navy?

matt-stephenson (approved):

domjacks (approved):

PR #190, opened 2021-01-20
Live edge change

matt-stephenson:
📺 What

Update the `isPlayingAtLiveEdge` API function.

> Tickets: N/A

🛠 How

Moves the `isPlayingAtLiveEdge` function to follow the same pattern as the stored `endOfStream` state which is returned through time updates and state changes. This is due to some devices not playing close enough to the live edge when started up for this function to report the playback as live.


jlks (approved):
Seems ok. Here was the original implementation for reference: https://github.com/bbc/bigscreen-player/pull/74

PR #192, opened 2021-02-05
Update isFirstManifest Logic to only compare Hosts 

ShiningTrapez:
📺 What

Currently, in some instances the first CDN is skipped as our check to determine whether it is the first CDN or not fails due to the `serviceLocation` URL from the manifest not always matching the CDN URLs passed in exactly.

> Tickets: IPLAYERTVV1-12007

🛠 How

Updates the logic for `isFirstManifest` to compare the domain of the serviceLocation instead of the full URL. This method uses a regex to pull out the hostname.

### DIFF
--- a/script-test/mediasourcestest.js
+++ b/script-test/mediasourcestest.js
@@ -241,7 +241,7 @@ require(
         });
       });
 
-      fdescribe('isFirstManifest', function () {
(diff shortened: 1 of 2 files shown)


Luke-Chatburn (approved):
Seems so complicated, but if it works 👍

PR #193, opened 2021-02-16
Subtitles Customisation API

jamierbower:
📺 What

- Customisable subtitles.
- Preview subtitles functionality.

[Add the jira ticket and/or the ADR to this PR]
> Tickets: [IPLAYERTVV1-11932](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-11932)

🛠 How

**Customisation API**:
- pass customisation options into imsc library. 

```
customisationOptions = { 
  size: Number,
  lineHeight: Number,
  backgroundColour: CSS String,
  fontFamily: String
}
```
Note: the `size` and `lineHeight` properties are on a scale from 0 - 1 (you can pass in bigger numbers but 1 is the authored size)

**Preview Subtitles API**:
- pass an example xml string, customisation options and a safe area for the preview to be rendered in. The preview can be removed from screen using a remove preview function.
- We expect the XML string to be compliant with IMSC 1

```safeArea = { left: Number } ```

Note: the `left` property is a percentage value between 0-100. We only take in to account left value for the safe area.

✅ Testing 
[MANDATORY for contributions being tested and released by the contributing team]

| ---- | ---- |

♿  Accessibility

adds the ability to increase the size of the subtitles.

 ✅ Acceptance criteria 

[MANDATORY for cross-team reviews]
* [X] [Add the acceptance criteria or some from the linked ticket number]
* [ ] [Mark correctly if it is covered in this PR or not]


matt-stephenson (approved):
Looks good, nice work.

I might have missed it, but I don't think I saw tearDown being tested on the Subtitles anymore, where as it was when it was in playerComponent.

It looks like the example string should be passed in as an XML formatted sting in the form needed by imsc.js. Was it considered to maybe pass in a simple string, and this code format that to a valid XML subs string?

jamierbower:
tearDown gets called in bsp's tearDown.

matt-stephenson:
> tearDown gets called in bsp's tearDown.

Yep, but I don't think there's a unit test to ensure that.

jamierbower:
good shout!

PR #197, opened 2021-03-02
BaseUrl failover for CDNs without a BaseUrl in the manifest (MSE Strategy)

jlks:
📺 What

Playback can failover with minimal user impact using the internal dash.js failover mechanism. Normally this would require an array of BaseUrls to be supplied in the Dash MPD, but this works around that mechanism and allows it to be done without.

🛠 How

- Ignore that there isn't a `BaseUrl` in the parsed manifest.
- Assume the content paths are the same across the supplied CDNs (see use of sources in `ManifestUpdater`).
- Refactor the code to be more readable.

✅ Testing 
| ---- | ---- |

### DIFF
--- a/script/manifest/manifestmodifier.js
+++ b/script/manifest/manifestmodifier.js
@@ -51,7 +51,6 @@ define('bigscreenplayer/manifest/manifestmodifier',
     function generateBaseUrls (manifest, sources) {
       var baseUrl = extractBaseUrl(manifest);
       var baseUrls = [];
-      // if (!baseUrl) return;
 
       function generateBaseUrl (source, priority, serviceLocation) {
         return {
--- a/script/manifest/manifestmodifier.js
+++ b/script/manifest/manifestmodifier.js
@@ -51,7 +51,7 @@ define('bigscreenplayer/manifest/manifestmodifier',
     function generateBaseUrls (manifest, sources) {
       var baseUrl = extractBaseUrl(manifest);
       var baseUrls = [];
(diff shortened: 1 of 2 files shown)


script/manifest/manifestmodifier.js
```diff
@@ -61,23 +72,36 @@ define('bigscreenplayer/manifest/manifestmodifier',
         };
       }
 
-      if (baseUrl.match(/^https?:\/\//)) {
+      function removeUnusedPeriodAttributes () {
+        if (manifest.Period && manifest.Period.BaseURL) delete manifest.Period.BaseURL;
+        if (manifest.Period && manifest.Period.BaseURL_asArray) delete manifest.Period.BaseURL_asArray;
```
domjacks, line 77: Will this mean that in the case that a manifest _did_ contain a `BaseURL_asArray` attribute, we'd always delete it?

jlks: It always gets removed from the `manifest.Period` but is always set on `manifest.BaseURL_asArray`.

domjacks: What happens if a manifest already has both a `manifest.Period` and `manifest.BaseURL_asArray`? Would we want to be deleting/overriding those?

domjacks (approved):

script/manifest/manifestmodifier.js
jlks: Picked up offline!

PR #201, opened 2021-03-23
Live subtitles

matt-stephenson:
📺 What

Adds ability to display subtitles for LIVE content

> Tickets: IPLAYERTVV1-11997

🛠 How

Moves loading of captions xml into subtitle containers (imsc and legacy). For live playback using imsc.js, individual fragments are downloaded when needed. The correct segment to render is determined and passed to imsc.js.

✅ Testing [Semi-optional]
[MANDATORY for contributions being tested and released by the contributing team]

| ---- | ---- |

♿  Accessibility [optional]

Adds the ability to display subtitles for live content.

### DIFF
--- a/script-test/subtitles/imscsubtitles.js
+++ b/script-test/subtitles/imscsubtitles.js
@@ -13,9 +13,7 @@ require(
       var fromXmlReturn;
       var mediaPlayer;
       var subtitles;
-      var stubCaptions = {
-        captionsUrl: 'http://stub-captions.test'
-      };
+      var stubCaptions;
 
       var loadUrlMock;
       var loadUrlStubResponseXml = '<?xml>';
@@ -23,6 +21,10 @@ require(
 
       beforeEach(function (done) {
         injector = new Squire();
+
+        stubCaptions = {
+          captionsUrl: 'http://stub-captions.test'
+        };
 
         mediaPlayer = jasmine.createSpyObj('mediaPlayer', ['getCurrentTime']);
         jasmine.clock().install();
@@ -327,6 +329,61 @@ require(
           expect(imscMock.fromXML).toHaveBeenCalledTimes(1);
           expect(imscMock.generateISD).toHaveBeenCalledTimes(1);
           expect(imscMock.renderHTML).toHaveBeenCalledTimes(1);
+        });
+      });
+
+      describe('Live subtitles', function () {
+        it('should load the first three segments with correct urls on instantiation', function () {
+          stubCaptions = {
+            captionsUrl: 'https://captions/$segment$.test',
+            segmentLength: 3.84
+          };
+
+          mediaPlayer.getCurrentTime.and.returnValue(10);
+          // 1614769200000 = Wednesday, 3 March 2021 11:00:00
+          subtitles = ImscSubtitles(mediaPlayer, stubCaptions, false, mockParentElement, {}, 1614769200000);
+
+          expect(loadUrlMock).toHaveBeenCalledWith('https://captions/420512815.test', jasmine.any(Object));
+          expect(loadUrlMock).toHaveBeenCalledWith('https://captions/420512816.test', jasmine.any(Object));
+          expect(loadUrlMock).toHaveBeenCalledWith('https://captions/420512817.test', jasmine.any(Object));
+        });
+
+        it('should load the fragment two segments ahead of current time at a frequency of segmentLength', function () {
+          stubCaptions = {
+            captionsUrl: 'https://captions/$segment$.test',
+            segmentLength: 3.84
+          };
+
+          mediaPlayer.getCurrentTime.and.returnValue(10);
+          // 1614769200000 = Wednesday, 3 March 2021 11:00:00
+          subtitles = ImscSubtitles(mediaPlayer, stubCaptions, false, mockParentElement, {}, 1614769200000);
+
+          loadUrlMock.calls.reset();
+          mediaPlayer.getCurrentTime.and.returnValue(13.84);
+          jasmine.clock().tick(3.84 * 1000);
+
+          // At 13.84 seconds, we should be loading the segment correseponding to 21.52 seconds
+          // 1614769221520 = Wednesday, 3 March 2021 11:00:21.52
+          expect(loadUrlMock).toHaveBeenCalledWith('https://captions/420512818.test', jasmine.any(Object));
+        });
+
+        it('should not load a fragment if fragments array already contains it', function () {
+          stubCaptions = {
+            captionsUrl: 'https://captions/$segment$.test',
+            segmentLength: 3.84
+          };
+
+          subtitles = ImscSubtitles(mediaPlayer, stubCaptions, false, mockParentElement, {}, 1614769200000);
+
+          loadUrlMock.calls.reset();
+          mediaPlayer.getCurrentTime.and.returnValue(13.84);
+          jasmine.clock().tick(3.84 * 1000);
+
+          expect(loadUrlMock).toHaveBeenCalledOnceWith('https://captions/420512818.test', jasmine.any(Object));
+
+          jasmine.clock().tick(3.84 * 1000);
+
+          expect(loadUrlMock).toHaveBeenCalledOnceWith('https://captions/420512818.test', jasmine.any(Object));
         });
       });
     });
--- a/script-test/subtitles/imscsubtitles.js
+++ b/script-test/subtitles/imscsubtitles.js
@@ -372,6 +372,7 @@ require(
       });
 
       describe('Live subtitles', function () {
+        var epochStartTimeMilliseconds = 1614769200000; // Wednesday, 3 March 2021 11:00:00
         beforeEach(function () {
           stubCaptions = {
             captionsUrl: 'https://captions/$segment$.test',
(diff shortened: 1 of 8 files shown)


script/subtitles/imscsubtitles.js
```diff
@@ -84,24 +146,43 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
+      function render (currentTime, xml) {
         removeCurrentSubtitlesElement();
 
         currentSubtitlesElement = document.createElement('div');
         currentSubtitlesElement.id = 'bsp_subtitles';
         currentSubtitlesElement.style.position = 'absolute';
         parentElement.appendChild(currentSubtitlesElement);
 
-        renderHTML(xml, currentTime, currentSubtitlesElement, imscRenderOpts, parentElement.clientHeight, parentElement.clientWidth);
+        return renderHTML(xml, currentTime, currentSubtitlesElement, imscRenderOpts, parentElement.clientHeight, parentElement.clientWidth);
```
domjacks, line 185: Any particular reason to return the result of the function here?

matt-stephenson: No, good spot. A hangover from when we did need to with the way we were checking if something had already been rendered.

script-test/subtitles/imscsubtitles.js
```diff
@@ -275,16 +371,369 @@ require(
         });
       });
 
-      describe('example rendering', function () {
-        it('should call fromXML, generate and render when renderExample is called', function () {
-          subtitles = ImscSubtitles(mediaPlayer, { xml: '', text: stubResponse }, false, mockParentElement, {});
-          imscMock.fromXML.calls.reset();
+      describe('Live subtitles', function () {
```
jlks, line 374: Does this make sense as a describe block given you're testing all of the functionality in here now?

script/bigscreenplayer.js
```diff
@@ -132,10 +132,11 @@ define('bigscreenplayer/bigscreenplayer',
 
         subtitles = Subtitles(
           playerComponent,
-          bigscreenPlayerData.media.captionsUrl,
+          bigscreenPlayerData.media.captions || { captionsUrl: bigscreenPlayerData.media.captionsUrl },
```
jlks, line 135: This looks like it extends the data model passed in to bigscreen player, maybe worth commenting and/or updating the docs

script/subtitles/imscsubtitles.js
```diff
@@ -4,28 +4,106 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
     'bigscreenplayer/plugins',
-    'bigscreenplayer/utils/playbackutils'
+    'bigscreenplayer/utils/playbackutils',
+    'bigscreenplayer/utils/loadurl'
   ],
-  function (IMSC, DOMHelpers, DebugTool, Plugins, Utils) {
+  function (IMSC, DOMHelpers, DebugTool, Plugins, Utils, LoadURL) {
     'use strict';
-    return function (mediaPlayer, response, autoStart, parentElement, defaultStyleOpts) {
+    return function (mediaPlayer, captions, autoStart, parentElement, defaultStyleOpts, windowStartTime) {
```
jlks, line 12: Suggestion: If you passed in mediasources you'd have access to `getWindowStartTime()` and also a reasonable place to put your segment calculation logic.

script/subtitles/imscsubtitles.js
```diff
@@ -4,28 +4,106 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
+          // This is to ensure when seeking to a point with no subtitles, don't leave previous subtitle displayed.
+          removeCurrentSubtitlesElement();
+        }
+
+        segmentsToLoad.forEach(function (segmentNumber) {
+          loadSegment(captions.captionsUrl, segmentNumber);
+        });
+      }
+
+      function calculateSegmentNumber (offset) {
```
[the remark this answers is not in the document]
Luke-Chatburn: Moved this into timeUtils and this also helped us spot that we don't need to keep calculating the current segment. However, don't think this has really given any value to the readability of the code as it was just a simple division.

script/bigscreenplayer.js
domjacks: We spoke about this briefly. I think ideally we would converge on a single data model ie. `captions` object. But given that would be a breaking API change, we can guard against it for now, document it and look to simplify in the future. 

Unless we think we can just update this now, do the docs and make it clear in semver the breaking change?

script/subtitles/imscsubtitles.js
```diff
@@ -138,10 +212,33 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
       }
 
+      function modifyStyling (xml) {
+        if (liveSubtitles) {
+          xml.head.styling.initials = defaultStyleOpts.initials;
+        }
+        return xml;
+      }
+
+      function timeIsValid (time) {
```
domjacks, line 222: Is this another candidate for timeutils? Or is it IMSC subs specific?

script/subtitles/imscsubtitles.js
```diff
@@ -4,28 +4,106 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
+          onLoad: function (responseXML, responseText, status) {
+            if (!responseXML && !liveSubtitles) {
+              DebugTool.info('Error: responseXML is invalid.');
+              Plugins.interface.onSubtitlesTransformError();
+              stop();
+              return;
+            }
+
+            try {
+              var xml = IMSC.fromXML(responseText.split(/<\?xml version=\"1.0\" encoding=\"UTF-8\"\?>/i)[1] || responseText);
```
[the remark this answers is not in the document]
Luke-Chatburn: Live subtitles contain more than just the XML in the response which is why we need to split. The XML header which we split on isn't needed to be considered valid XML.

script/subtitles/imscsubtitles.js
```diff
@@ -84,24 +146,43 @@ define('bigscreenplayer/subtitles/imscsubtitles',
...
+        if (segment) {
+          DebugTool.keyValue({key: 'segmentToRender', value: 'true'});
+          render(currentTime, segment.xml);
+        } else {
+          DebugTool.keyValue({key: 'segmentToRender', value: 'false'});
         }
       }
 
-      function render (currentTime) {
+      function getSegmentToRender (currentTime) {
```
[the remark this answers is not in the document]
Luke-Chatburn: We made a decision to use the term segment instead of fragment.

script/bigscreenplayer.js
Luke-Chatburn: Yes, we plan on updating the docs for BSP with live subtitle information. We intentionally built it this way so it worked with the old method and VOD subs wouldn't require testing for our earlier upstream changes. My plan was to create more work to be done after that was go to one data model. I didn't expect us to make so many VOD changes as part of this work but I still feel we made the right decision to break up the work this way.

script/subtitles/imscsubtitles.js
Luke-Chatburn: Since currently subtitles is only interested in windowStartTime I feel it makes more sense to just pass it in directly.

script/subtitles/imscsubtitles.js
Luke-Chatburn: timeIsValid in this context is very live subtitle specific and is just our way of protecting us from requesting segments that are not needed when the player time isn't correct yet.

domjacks: 👍

script/subtitles/imscsubtitles.js
```diff
@@ -138,10 +219,33 @@ define('bigscreenplayer/subtitles/imscsubtitles',
         }
       }
 
+      function modifyStyling (xml) {
+        if (liveSubtitles) {
+          xml.head.styling.initials = defaultStyleOpts.initials;
```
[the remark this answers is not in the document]
Luke-Chatburn: Is this not protected by the try catch in loading of segments?

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
domjacks: Is there an industry standard naming convention? I know in the BBC guidelines we tend to use fragment. What does DASH or the TextTrack API use for example?

It could be useful to be able to distinguish between media segments and subtitle fragments 🤷

script/bigscreenplayer.js
domjacks: Yep, I agree @Luke-Chatburn. I think this was a sensible approach 🚀

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
Luke-Chatburn: No idea. We originally had segment/fragment everywhere and it was more confusing, fragmented / segmented mean the same thing to me. Since we fetch captions using the same method as media it made more sense to use segment.

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
domjacks: I guess it depends what we want to do in the situation that the values don't exist. Do we want to no show subs, or show subs in an unstyled manner.

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
domjacks: As long as we converge on one or the other and we're not mixing within the same context (ie. always referenced as segment in this code) then I guess it's fine for now. We can always agree on better naming another time. 

Unless there are strong thoughts otherwise?

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
matt-stephenson: So this will currently show subs without this styling override. As part of future work we could emit another event through the plugins so that clients could monitor errors in this case.

script/subtitles/imscsubtitles.js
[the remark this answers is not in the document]
domjacks: My gut feeling says we shouldn't mutate the XML object by stripping the head off. 

However if IMSC accepts it then it's likely fine. Hoping we don't get tripped up by this in the future though if we were to use the XML for something else further in the code.

The reason I say this is because removing the encoding declaration _could_ be an issue. But I imagine parsers autodetect this stuff so it should be fine.

domjacks (approved):

PR #204, opened 2021-04-07
Subtitle CDN failover

jamierbower:
📺 What

Introduce CDN failover capability for subtitles.

> Tickets: [IPLAYERTVV1-11794](https://jira.dev.bbc.co.uk/browse/IPLAYERTVV1-11794)

🛠 How

- initialise BigscreenPlayer with a captions array.
- failover to subsequent sources if loading the subtitles xml is unsuccessful.

✅ Testing 

| ---- | ---- |

- [x] Subtitles continue to display if a cdn fails
- [x] Subtitles failover for VOD
- [x] Subtitles failover for Simulcast
- [x] Video failover not affected
- [x] Subtitles remain in sync after failover
- [x] Legacy and IMSC subtitles to have failover functionality

### DIFF
--- a/development/index.js
+++ b/development/index.js
@@ -15,7 +15,7 @@ require (['bigscreenplayer/bigscreenplayer'], function(BigscreenPlayer){
   let minimalData = {
     initialPlaybackTime: 30,
     media: {
-      captionsUrl: '',
+      captions: [],
         type: 'application/dash+xml',
         mimeType: 'video/mp4',
         kind: 'video',
--- a/development/index.js
+++ b/development/index.js
@@ -10,7 +10,7 @@ require (['bigscreenplayer/bigscreenplayer'], function(BigscreenPlayer){
   playbackElement.style.width = '1280px';
 
   let windowType = 'staticWindow';
-  let enableSubtitles = true;
+  let enableSubtitles = false;
   
   let minimalData = {
     initialPlaybackTime: 30,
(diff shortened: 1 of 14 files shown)


script/bigscreenplayer.js
```diff
@@ -132,11 +132,10 @@ define('bigscreenplayer/bigscreenplayer',
 
         subtitles = Subtitles(
           playerComponent,
-          bigscreenPlayerData.media.captions || { captionsUrl: bigscreenPlayerData.media.captionsUrl },
```
jlks, line 135: Can't you maintain backwards compatibility with the old API? This shouldn't be too hard to do

jamierbower: We could, I'm not sure we need to as media selector will always be returning a captions object now

development/index.js
```diff
@@ -15,20 +15,18 @@ require (['bigscreenplayer/bigscreenplayer'], function(BigscreenPlayer){
...
-          {
-            // Content from DASH IF testing assests (used in their reference player)
-            // https://reference.dashif.org/dash.js/v2.9.2/samples/dash-if-reference-player/index.htm
-            url: 'https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd',
-            cdn: 'dash.akamaized.net'
-          }
-        ]
-      }
-    };
+      captions: [],
```
domjacks, line 18: Maybe not doable, but is there a supplied example vid that _does_ have subtitles we could use?

jamierbower: I'll look into it!

matt-stephenson (approved):

PR #208, opened 2021-04-21
Report CDN as an event through subtitle error plugins

matt-stephenson:
📺 What

Report current CDN through subtitle error plugins where appropriate (load, timeout, xml errors)

> Tickets: IPLAYERTVV1-11795

🛠 How

Expose current subtitle cdn on `mediasources`. Move subtitle error plugin into media sources and fire whenever failover is attempted, with a severity level of FAILOVER or FATAL, depending on whether there are further cdns to attempt. 

✅ Testing [Semi-optional]

| ---- | ---- |

### DIFF
--- a/script/mediasources.js
+++ b/script/mediasources.js
@@ -197,6 +197,14 @@ define('bigscreenplayer/mediasources',
         return subtitlesRequestTimeout;
       }
 
+      function getCurrentSubtitlesCdn () {
+        if (subtitlesSources.length > 0) {
+          return subtitlesSources[0].cdn;
+        }
+
+        return undefined;
+      }
+
       function availableUrls () {
         return mediaSources.map(function (mediaSource) {
           return mediaSource.url;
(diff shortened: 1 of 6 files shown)


script/subtitles/legacysubtitles.js
```diff
@@ -33,11 +33,11 @@ define(
             onError: function (statusCode) {
               var errorCase = function () { Plugins.interface.onSubtitlesLoadError({status: statusCode}); };
               DebugTool.info('Error loading subtitles data: ' + statusCode);
-              mediaSources.failoverSubtitles(loadSubtitles, errorCase);
+              mediaSources.failoverSubtitles(loadSubtitles, errorCase, statusCode);
```
jamierbower, line 36: This means you will be calling onSubtitlesLoadError, once when you call the errorCase callback, and once in mediaSources failoverSubtitles

matt-stephenson: Good spot, I sorted this for imsc, but not legacy. Thanks.

script/mediasources.js
```diff
@@ -197,6 +199,14 @@ define('bigscreenplayer/mediasources',
         return subtitlesRequestTimeout;
       }
 
+      function getCurrentSubtitlesCdn () {
+        if (subtitlesSources.length > 0) {
+          return subtitlesSources[0].cdn;
+        }
+
+        return undefined;
```
domjacks, line 207: Is it better to just not send the `cdn` field if it's not there for some reason? Or is it a required field and therefore `undefined` is fine?

matt-stephenson: Was following the pattern from the other functions to be honest.

jamierbower (approved):

domjacks (approved):

Issue #195, opened 2021-02-23
Add top, bottom and right percentage to `safePosition` when rendering an example Subtitle


domjacks:
Currently, when passing a `safePosition` into the `renderExample` function, you can pass it a `left` value.

A better API would allow all 4 percentage values (`top`, `bottom`, and `right`) to be set so that other clients can position the example subtitle wherever they want it. 

```
var safePosititon = { top: 0, right: 0, left: 30, bottom: 20 };
bigscreenPlayer.renderSubtitleExample(exampleUrl, customStyleObj, safePosititon);
```

PR #209, opened 2021-04-21
Subtitle Customisation: Also support `top`, `bottom` and `right` percentage on`SafePosition`

tjenkinson:
📺 What
Also support `top`, `bottom` and `right` percentage on `SafePosition` for the subtitle preview.

This will give the UI team more flexibility on the position of the menu. I.e. it doesn't need to be fixed to the left of the screen.

Follow up to https://github.com/bbc/bigscreen-player/pull/193
Fixes https://github.com/bbc/bigscreen-player/issues/195

> Tickets: IPLAYERTVV1-12243

🛠 How
- Also handle `top`, `right` and `bottom` (defaulting them to 0 to preserve backwards compatibilty)
- Add some tests that the preview dom element is correct.

✅ Testing [Semi-optional]
Probably good enough to get this version in an app branch and check that the dom element looks correct?

| ---- | ---- |

The dom element looks like 
image (but with `px` values not `%`)  in the app and takes up the correct space in the UI.

### DIFF
--- a/script/subtitles/imscsubtitles.js
+++ b/script/subtitles/imscsubtitles.js
@@ -221,17 +221,15 @@ define('bigscreenplayer/subtitles/imscsubtitles',
         exampleSubtitlesElement.id = 'subtitlesPreview';
         exampleSubtitlesElement.style.position = 'absolute';
 
-        var renderWidth = parentElement.clientWidth;
-        if (safePosition) {
-          exampleSubtitlesElement.style.left = safePosition.left + '%';
-
-          var leftPixels = parentElement.clientWidth * (safePosition.left / 100);
-          renderWidth = parentElement.clientWidth - leftPixels;
-        }
+        safePosition = safePosition || {};
+        exampleSubtitlesElement.style.top = (safePosition.top || 0) + '%';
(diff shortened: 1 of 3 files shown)


script/subtitles/imscsubtitles.js
```diff
@@ -221,17 +221,15 @@ define('bigscreenplayer/subtitles/imscsubtitles',
         exampleSubtitlesElement.id = 'subtitlesPreview';
         exampleSubtitlesElement.style.position = 'absolute';
 
-        var renderWidth = parentElement.clientWidth;
-        if (safePosition) {
-          exampleSubtitlesElement.style.left = safePosition.left + '%';
-
-          var leftPixels = parentElement.clientWidth * (safePosition.left / 100);
-          renderWidth = parentElement.clientWidth - leftPixels;
```
tjenkinson, line 229: ah this is mistake. bringing back

tjenkinson: hmm ok needs a bit more work for actually. will make this draft for now

domjacks (approved):
LGTM

PR #210, opened 2021-04-27
Legacy subs - Only load subtitle urls when required

Luke-Chatburn:
📺 What

* Legacy subs - Only load sub urls when required
* IMSC subs - Don't keep loading subtitles for VOD on start/stop

> Tickets: IPLAYERTVV1-11796

🛠 How

* Put the current initial loading of subtitles behind an autoStart check. (Similar to IMSC implementation)
* if autoStart is false and subtitles are started mid stream. Load subs before starting.


jamierbower (approved):
Good stuff!

domjacks (approved):

PR #211, opened 2021-04-27
Fix IMSC subs tests when running Jasmine in local browser

domjacks:
📺 What

Updates some tests for IMCS subtitles to check the HTML element style properties, instead of the rendered HTML string. 

This stops the problem where browsers may interpret CSS styles and modify them.

In this case, Chrome was changing:
```
<div id="subtitlesPreview" style="position: absolute; top: 1px; right: 1px; bottom: 1px; left: 1px;"></div>
```
to:
```
<div id="subtitlesPreview" style="position: absolute; inset: 1px;"></div>
```

Notice the individual locations were replaced by a single `inset` value.

This only matters when you're not running the Jasmine test runner using the headless Chrome provided by `grunt-contrib-jasmine` package and instead running it in a more up-to-date local browser on your desktop.

🛠 How

Assert against the `subsElement` object and it's `style` properties directly.

### DIFF
--- a/script-test/subtitles/imscsubtitles.js
+++ b/script-test/subtitles/imscsubtitles.js
@@ -199,11 +199,11 @@ require(
         it('should call renderHTML with a preview element with the correct structure when no position info', function () {
           subtitles = ImscSubtitles(mediaPlayer, false, mockParentElement, mockMediaSources, undefined);
 
-          var exampleHtml = null;
+          var exampleSubsElement = null;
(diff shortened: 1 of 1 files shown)


tjenkinson (approved):
Oops LGTM!

I’ve done something similar before but with jest snapshots which I think uses jsdom, which is deterministic.

Luke-Chatburn (approved):

PR #215, opened 2021-05-18
Ability to set/get playback rate

matt-stephenson:
📺 What

Add ability to set the playback rate

> Tickets: IPLAYERTVV1-12315

🛠 How

Add api methods for get and set.

When setting:
- MSE strategy - will call method on dash.js
- Native strategy - will set the playback rate directly on the mediaElement (only for the html5 media player - so will **not** currently work for live or other players such as cehtml using native strategy)
- Basic strategy - will set playback rate directly

When getting - for native strategy, attempting to getPlaybackRate on anything other than the standard html5 mediaPlayer will result in a return value of 1.0. This is because you cannot _**set**_ the rate for these non-standard players.

### DIFF
--- a/script-test/bigscreenplayertest.js
+++ b/script-test/bigscreenplayertest.js
@@ -126,7 +126,7 @@ require(
         var mockDebugTool = jasmine.createSpyObj('mockDebugTool', ['apicall', 'time', 'event', 'keyValue', 'tearDown', 'setRootElement']);
         mockPlayerComponentInstance = jasmine.createSpyObj('playerComponentMock', [
           'play', 'pause', 'isEnded', 'isPaused', 'setCurrentTime', 'getCurrentTime', 'getDuration', 'getSeekableRange',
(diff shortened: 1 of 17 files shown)


domjacks (approved):
🚀

Luke-Chatburn:
If this is only supported for a subset of strategies do we need to add a support option?

Do we need to add an initial playback rate to the initial data passed into BSP? i.e playback rate is stored and always wants to be 2x until changed.

matt-stephenson:
@Luke-Chatburn - I'm guessing you mean an `isPlaybackRateSupported()` type of function exposed, so UI's can decide whether to provide the option or not? If so, maybe in the future yes, this is a first pass, get it working at attempt. If you mean, does it need config so it's not attempted at all, then no, it should guard around unsupported strategies and just not affect the rate if you try.

Again, with the initial speed thing, this would also be in the category of 'one for the future'.

Luke-Chatburn (approved):

PR #220, opened 2021-06-14
Allow dashjs configuration for msestrategy

Luke-Chatburn:
📺 What

Allow dashjs streaming settings to be customised. 

> Tickets: IPLAYERTVV1-11998

🛠 How

* The media block now accepts an mseCustomisation object. Which will be passed down into the strategy.
* Change how our utils object merging works to allow deep objects.

### DIFF
--- a/script/playercomponent.js
+++ b/script/playercomponent.js
@@ -30,7 +30,7 @@ define(
         mediaKind,
         playbackElement,
         bigscreenPlayerData.media.isUHD,
-        bigscreenPlayerData.media.mseCustomisation
+        bigscreenPlayerData.media.playerSettings
       );
 
       playbackStrategy.addEventCallback(this, eventCallback);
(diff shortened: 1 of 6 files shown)


matt-stephenson (approved):

script/playercomponent.js
```diff
@@ -29,7 +29,8 @@ define(
         windowType,
         mediaKind,
         playbackElement,
-        bigscreenPlayerData.media.isUHD
+        bigscreenPlayerData.media.isUHD,
+        bigscreenPlayerData.media.mseCustomisation
```
domjacks, line 33: Is there a chance we'd want to set configuration options for non-MSE strategies?

Luke-Chatburn: I would honestly say that is never going to happen but down to change the name to something more generic. playerSettings?

domjacks: Yeah I like `playerSettings`.

It's just on the off chance we want to influence some configuration across multiple or a different strategy. Like you said, likely to never happen but might be worth making generic now in case.

domjacks (approved):

PR #226, opened 2021-07-01
Readme fix

RelentlessExploration:
📺 What

Updated README, improved code correctness.

🛠 How

First of all, I wrapped the `MediaStates` in `README.md` with backticks for good measure because the media states are code, thus they should be formatted properly as code as well, adding backticks makes it look more accurate, meaningful, and purposeful. Also added commas in objects as this is a syntax requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,11 @@ require(
             cdn: 'cdn1' // Displayed by Debug Tool
           }, {
             url: 'https://www.somelovelycaptionsurl2.com/captions/$segment$',
-            segmentLength: 3.84 
-            cdn: 'cdn1' 
+            segmentLength: 3.84
+            cdn: 'cdn1'
           },
         ],
-        captionsUrl: 'https://www.somelovelycaptionsurl.com/captions/', // NB This parameter is being deprecated in favour of the captions array shown above. 
+        captionsUrl: 'https://www.somelovelycaptionsurl.com/captions/', // NB This parameter is being deprecated in favour of the captions array shown above.
         subtitlesRequestTimeout: 5000, // Optional override for the XHR timeout on sidecar loaded subtitles
         subtitleCustomisation: {
           size: 0.75,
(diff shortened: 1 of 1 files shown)


domjacks (approved):
Thank for very much for your contribution @RelentlessExploration 😄 . We will look to merge this ASAP 🚀

RelentlessExploration:
Thanks, hope this helps, at least in those very little details!

README.md
```diff
@@ -73,15 +73,15 @@ require(
...
-            segmentLength: 3.84 // Required to calculate live subtitle segment to fetch & live subtitle URL.
-            cdn: 'cdn1' // Displayed by Debug Tool
+            segmentLength: 3.84, // Required to calculate live subtitle segment to fetch & live subtitle URL.
+            cdn: 'cdn1', // Displayed by Debug Tool
           }, {
             url: 'https://www.somelovelycaptionsurl2.com/captions/$segment$',
-            segmentLength: 3.84 
-            cdn: 'cdn1' 
+            segmentLength: 3.84,
+            cdn: 'cdn1',
```
domjacks, line 81: Just noticed @RelentlessExploration that there's trailing commas in here. Although not illegal from a ES5 point of view, we do have an eslint rule to not allow them so for consistency with our codebase it would be great if you could remove any trailling commas 😄 Thanks

RelentlessExploration: Will do, thanks!

RelentlessExploration:
Done.

domjacks:
Another tiny observation @RelentlessExploration, which you can choose to ignore if you'd like, a bit further down there's the 'Creating a plugin' section which lists some interface functions. These could also be surrounded by backticks if you wanted 😄

RelentlessExploration:
I hope this is what you are referring to: 
Screen Shot 2021-07-05 at 2 13 52 PM

domjacks:
🚀

PR #231, opened 2021-07-19
CDN Failover Recovery

jlks:
📺 What

After 2 minutes, bring a CDN that has failed back in to service.

> Tickets: IPLAYERTVV1-12466

🛠 How

- Add a timeout to restore a failed CDN back in to service.
- Make the timeout configurable through the `playerSettings.failoverResetTime`.
- Add an optional sorting algorithm to sort the list of CDNs after failover (e.g priority, weighting, other criteria based on the `media.urls[].url` object).

Note:
- This will always add the failed CDN back in to service. However, if a device fails all CDNs before a failed one is added back, playback and the failover process will fail as before.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ buildVersion.js
 buildExternal.js
 webpack.config.js
 /development/index.js
+/development/
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,4 +5,3 @@ buildVersion.js
 buildExternal.js
 webpack.config.js
 /development/index.js
-/development/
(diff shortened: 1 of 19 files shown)


.eslintignore
```diff
@@ -5,3 +5,4 @@ buildVersion.js
 buildExternal.js
 webpack.config.js
 /development/index.js
+/development/
```
domjacks, line 8: Do you need the line above now?

script/playbackstrategy/msestrategy.js
```diff
@@ -332,7 +344,7 @@ define('bigscreenplayer/playbackstrategy/msestrategy',
         mediaPlayer = dashjs.MediaPlayer().create();
         var playerSettings = Utils.merge({
           debug: {
-            logLevel: 2
+            logLevel: 4
```
Luke-Chatburn, line 347: I assume you don't want this all the time.

jlks: Good spot, thought I'd solved that with an earlier commit!

Luke-Chatburn (approved):

domjacks (approved):

PR #237, opened 2021-09-01
Make subtitle header XML split more general

domjacks:
📺 What

Allows the XML declaration to contain more generic content including the usage of both single and double quotation marks 

> Tickets: IPLAYERTVV1-12583

🛠 How

Update the regex used to match the declaration in the `.split` function when preparing the document for IMSC.

 ✅ Acceptance criteria 

* [ ] XML documents with varying XML declarations can be used to provide subtitles

### DIFF
--- a/script-test/subtitles/imscsubtitles.js
+++ b/script-test/subtitles/imscsubtitles.js
@@ -283,7 +283,7 @@ require(
           loadUrlStubResponseText = '<?xml version="1.0" encoding="utf-8" extra property="something"?><tt xmlns="http://www.w3.org/ns/ttml"></tt>';
           subtitles = ImscSubtitles(mediaPlayer, true, mockParentElement, mockMediaSources, {});
 
-          expect(imscMock.fromXML).toHaveBeenCalledWith(loadUrlStubResponseText);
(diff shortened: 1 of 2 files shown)


script-test/subtitles/imscsubtitles.js
```diff
@@ -283,7 +283,7 @@ require(
           loadUrlStubResponseText = '<?xml version="1.0" encoding="utf-8" extra property="something"?><tt xmlns="http://www.w3.org/ns/ttml"></tt>';
           subtitles = ImscSubtitles(mediaPlayer, true, mockParentElement, mockMediaSources, {});
 
-          expect(imscMock.fromXML).toHaveBeenCalledWith(loadUrlStubResponseText);
+          expect(imscMock.fromXML).toHaveBeenCalledWith('<tt xmlns="http://www.w3.org/ns/ttml"></tt>');
```
domjacks, line 286: This test needed to be updated as previously the regex wasn't matching (because of the extra property bit) meaning the split did not remove the XML declaration. Now it does match, it will be removed when doing a `.split`

ShiningTrapez (approved):
LGTM

PR #255, opened 2022-05-26
Handle `play()` promise and ignore `AbortError`

tjenkinson:
📺 What

Handle `play()` promise and ignore `AbortError`.

This error happens when a `play()` request is aborted due to a `pause()` call, which I think isn't really an error we need to handle.

Without this change you get an `unhandledpromiserejection` event, which is ending up causing noise in telemetry.

🛠 How
Provide a helper that takes a play promise and ignores the error if it's `AbortError`, otherwise rethrows it given it's unexpected. It also handles there being no promise, as not all devices will provide one.

✅ Testing
I think we just need to check on a few devices that it still works.

| ---- | ---- |


Luke-Chatburn (approved):
Looks good

PR #262, opened 2022-06-09
Protect all interactions with Seekable property when using the caching override.

Luke-Chatburn:
📺 What

Adds additional caching protection to the seekable range properties.

> Tickets: IPLAYERTVV1-12900

🛠 How

* Encapsulates the calls to all the seekable range properties inside a `getElementSeekableRange` so `getCachedSeekableRange` can now use this and won't interact with the `mediaElement.seekable` more than every 250ms.

### DIFF
--- a/src/playbackstrategy/modifiers/html5.js
+++ b/src/playbackstrategy/modifiers/html5.js
@@ -267,7 +267,6 @@ function Html5 () {
   }
 
   function setSentinels (sentinels) {
-    disableSeekSentinel = true
     if (disableSentinels) { return }
 
     clearSentinels()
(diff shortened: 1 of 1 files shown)


src/playbackstrategy/modifiers/html5.js
```diff
@@ -373,7 +366,14 @@ function Html5 () {
         }
       }
     }
-    return undefined
```
matt-stephenson, line 376: Just wondering if leaving out this return undefined was deliberate?

Luke-Chatburn: Yes, the function will return undefined.

PR #264, opened 2022-07-08
Update node

domjacks:
📺 What

Ensure Node v16 is being used

🛠 How

Update `.nvmrc` to 16 and replace `package-lock.json`


ShiningTrapez (approved):
Finally, we can move past npm@6

PR #265, opened 2022-07-11
Emit media element errors

matt-stephenson:
📺 What

Emit a MediaError code (see [here](https://developer.mozilla.org/en-US/docs/Web/API/MediaError/code)) when state changes to FATAL_ERROR or through the plugin interface when recovering from a CDN failover.

> Tickets: IPLAYERTVV1-13229

🛠 How

Native and Basic strategies listen directly to the media element, and publish a received error code/message, or a default of `0` and `unknown`.

MSE strategy propagates errors raised by dash.js. This can include the original four MediaError errors, but also an extended set of codes from dash.js (e.g. when segments cannot be found, or the manifest cannot be loaded).

✅ Testing [Semi-optional]

| ---- | ---- |


ShiningTrapez (approved):

src/playercomponent.js
```diff
@@ -190,46 +190,48 @@ function PlayerComponent (playbackElement, bigscreenPlayerData, mediaSources, wi
...
         fatalErrorTimeout = null
         fatalError = true
-        attemptCdnFailover(false)
+        attemptCdnFailover(false, mediaError)
       }, 5000)
     }
   }
 
-  function attemptCdnFailover (bufferingTimeoutError) {
+  function attemptCdnFailover (bufferingTimeoutError, mediaError) {
```
matt-stephenson, line 224: This could use the mediaError code instead of the separate `bufferingTimeoutError` argument.

src/playbackstrategy/basicstrategy.js
```diff
@@ -121,7 +121,11 @@ function BasicStrategy (mediaSources, windowType, mediaKind, playbackElement, is
   }
 
   function onError (_event) {
-    publishError()
+    let mediaError = {
+      code: mediaElement.error?.code || 0,
```
matt-stephenson, line 125: Remove optional chaining for safety on some devices?

src/playbackstrategy/modifiers/html5.js
```diff
@@ -293,8 +294,9 @@ function Html5 () {
     }, 1100)
   }
 
-  function reportError (_errorMessage) {
-    emitEvent(MediaPlayerBase.EVENT.ERROR)
+  function reportError (errorString, mediaError) {
+    DebugTool.info('HTML5 Media Player error: ' + errorString)
```
domjacks, line 298: 👍

src/playbackstrategy/msestrategy.js
```diff
@@ -243,15 +251,19 @@ function MSEStrategy (mediaSources, windowType, mediaKind, playbackElement, isUH
   function onBaseUrlSelected (event) {
     const failoverInfo = {
       errorMessage: 'download',
-      isBufferingTimeoutError: false
+      isBufferingTimeoutError: false,
+      code: lastError && lastError.code,
+      message: lastError && lastError.message
```
domjacks, line 256: We now have `errorMessage` and `message` in the same object. Is there a way we can distinguish between these two better?

domjacks (approved):
Looks fine. I think we should have a think about error handling in general in the playerComponent. I'm not sure how much it makes sense anymore with the addition of mediaErrors being bubbled up.

src/playbackstrategy/msestrategy.js
matt-stephenson: Good spot. Digging further, errorMessage is unused, so have removed that too here.

PR #273, opened 2022-08-18
Handle and report manifest errors and provide a fallback

jlks:
📺 What

Currently there is no way to know manifest parsing has failed. This PR provides a plugin hook `onManifestParseError` and explicitly defines the fallback behaviour of ManifestParser.

> Tickets: N/A

🛠 How

- Add `onManifestParseError` to the plugin interface.
- Change returning undefined from `convertEpochMsToVideoTimeSeconds` and `convertVideoTimeSecondsToEpochMs` to return null.

### DIFF
--- a/src/manifest/manifestparser.js
+++ b/src/manifest/manifestparser.js
@@ -1,6 +1,7 @@
 import TimeUtils from './../utils/timeutils'
 import DebugTool from '../debugger/debugtool'
 import Plugins from '../plugins'
+import pluginenums from '../pluginenums'
 
 function parseMPD (manifest, dateWithOffset) {
   try {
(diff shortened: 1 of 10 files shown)


src/pluginenums.js
```diff
@@ -10,6 +10,7 @@ export default {
     ERROR: 'error'
   },
   ERROR_CODES: {
+    MANIFEST_PARSE: 7,
     BUFFERING_TIMEOUT: 8,
     MANIFEST: 9
```
matt-stephenson, line 15: Could we change this to MANIFEST_LOAD?

src/manifest/manifestparser.js
```diff
@@ -26,26 +29,34 @@ function parseMPD (manifest, dateWithOffset) {
         correction: timeCorrection
       }
     } else {
-      return { error: 'Error parsing DASH manifest attributes' }
+      throw new Error('manifest-dash-attributes-parse-error')
     }
   } catch (e) {
-    return { error: 'Error parsing DASH manifest' }
+    const error = new Error(e.message || 'manifest-dash-parse-error')
```
Luke-Chatburn, line 35: Will we not get a different message here if it actually fails on the try/catch? Do we not want it to always be either 'manifest-dash-parse-error' or 'manifest-dash-attributes-parse-error'

Also, we only ever report the single code and we have multiple paths of errors here? do we want to report this extra information?

src/pluginenums.js
matt-stephenson: Sorry, maybe GitHub formatting, I mean the one on line 15, that's currently just MANIFEST.

Luke-Chatburn (approved):

src/manifest/manifestparser.js
jlks: Yeah you end up with the different message but the same code, both of which are reported to the plugin. It is a bit annoying having to rely on the error throwing upwards.

PR #276, opened 2022-11-09
Set up linting and lint and formatting

eirikbjornr:
📺 What

Set up opinionated linting and formatting with eslint and prettier.

🛠 How

Prettier and ESLint runs when you commit to lint and format the commited code. If code fails during linting the commit fails as well.

See `scripts/format-staged.sh` and `lint-staged.sh` for implementation.

Running an `eslint --fix` and `prettier --write` on the codebase is left as a `todo` after this is merged because we would want the output to be tested. We want to agree on a set of linting and formatting rules before we re-lint and re-format the code on a large scale.

This lets us gradually lint our code to a new standard instead of creating a gigantic PR. It should also keep the conversation around code style and quality flowing for a while until we agree on a complete set of rules. (...aka forever?? 👀)


eirikbjornr:
TODO: only lint staged files

ShiningTrapez (approved):

ShiningTrapez (approved):

PR #278, opened 2022-12-01
MAJOR! Refactor bsp to class

eirikbjornr:
📺 What

Refactor BigscreenPlayer to a Javascript `Class`.

🛠 How

Changed BigscreenPlayer from function to `Class`, updating tests accordingly.

💡 Why

A step closer to implementing f.ex. the `EventTarget` API. This removes the need for registered callbacks and plugins in their current usage. This feature (+ `@deprecated`-ing all `registerCb` and `plugins`) would be a good next step.

✅ Testing

TBD

 ✅ Acceptance criteria
 
 TBD


ShiningTrapez (commented):
LGTM, Like the use of extremely modern JS like class fields

PR #280, opened 2023-03-03
Set Dash.js log level depending on DebugTool log level

domjacks:
📺 What

Set the Dash.js [debug player settings](http://cdn.dashjs.org/latest/jsdoc/module-Settings.html#~DebugSettings__anchor) depending on the overall library log level. 

🛠 How

- Adds new `getLogLevel()` function to DebugTool and player API
- Sets Dash.js player settings for debug based on the LOG_LEVEL set

INFO = `{ logLevel: 2}` dashjs.Debug.LOG_LEVEL_ERROR - Log error messages.
VERBOSE = `{ logLevel: 5, dispatchEvent: true}` // dashjs.Debug.LOG_LEVEL_DEBUG - Log debug messages.

### DIFF
--- a/src/bigscreenplayer.js
+++ b/src/bigscreenplayer.js
@@ -609,6 +609,12 @@ function BigscreenPlayer () {
 
     /**
      * @function
+     * @return {string} - currently set log level @see getLogLevels
+     */
+    getLogLevel: () => DebugTool.getLogLevel,
+
+    /**
+     * @function
      * @param logLevel -  log level to display @see getLogLevels
      */
     setLogLevel: DebugTool.setLogLevel,
--- a/src/bigscreenplayer.js
+++ b/src/bigscreenplayer.js
@@ -47,7 +47,7 @@ function BigscreenPlayer () {
       DebugTool.time(evt.data.currentTime)
       callCallbacks(timeUpdateCallbacks, {
         currentTime: evt.data.currentTime,
-        endOfStream: endOfStream
+        endOfStream
       })
     } else {
       let stateObject = { state: evt.data.state }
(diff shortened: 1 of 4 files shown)


src/bigscreenplayer.js
```diff
@@ -607,6 +607,12 @@ function BigscreenPlayer () {
      */
     getLogLevels: () => DebugTool.logLevels,
 
+    /**
+     * @function
+     * @return {string} - currently set log level @see getLogLevels
+     */
+    getLogLevel: () => DebugTool.getLogLevel,
```
eirikbjornr, line 614: This should be `DebugTool.getLogLevel()`

eirikbjornr (changes requested):
LGTM! Only thing I've spotted – Think you're returning the `getLogLevel` function instead of the current log level in `bigscreenplayer`

src/bigscreenplayer.js
domjacks: In fact, I think it should just be `getLogLevel: DebugTool.getLogLevel`

To be honest, I'm not even sure it's valuable adding it to the top level API.

eirikbjornr:
Just realised tyou prolly hadn't opened for review yet 😅 I saw the PR in our Slack channel and autopilot assumed you were requesting review sorry!

src/bigscreenplayer.js
eirikbjornr: That's a fair call too!

Is your thinking to export `DebugTools` in `main.js` instead?

domjacks: No, I'm just not sure if people will actually use the API outside of the BSP codebase itself to get the log level. I'll leave it just in case, not harming anything 😄

domjacks:
One problem with this is that you will need to set VERBOSE before starting playback using Dash.js. This mechanism doesn't currently update player settings in Dash.js should the log level be set to VERBOSE mid-stream. 

Options:

- We could figure out a way to update the player settings (using dash.js `updateSettings()` function) whenever setLogLevel is called
- Not do this PR and just rely on people knowing that in order to get Dash.js debug logs, you need to set log level to VERBOSE and set the player settings before starting the stream
- Do this PR but people will need to know that if you set log level after setting up the media player, then the dash.js log output won't be logged.

PR #282, opened 2023-03-14
Decouple the logic that handles subtitles delivered as segments from stream type

eirikbjornr:
📺 What

Decouple the logic that handles subtitles delivered as segments from stream type.

> Tickets: IPLAYERTVV1-13683

🛠 How

1. Fetch the seek window start time off the manifest when subtitles are delivered as segments by checking for a segment template in the subtitles url.
2. Fetch multiple segments when there is a segment template in the subtitles url.


ShiningTrapez (approved):
LGTM!

ShiningTrapez (approved):

PR #284, opened 2023-03-27
Don't attempt to load Legacy Subs on Live

ShiningTrapez:
📺 What
Don't attempt to load Legacy Subs on Live

> Tickets: BADGERS-158

🛠 How
With the new Whole/Segmented distinction, we can simply check that subtitles are available before making the request


src/subtitles/subtitles.js
```diff
@@ -8,14 +8,20 @@ function Subtitles(mediaPlayer, autoStart, playbackElement, defaultStyleOpts, me
...
+        .then(({ default: LegacySubtitles }) => {
+          subtitlesContainer = LegacySubtitles(mediaPlayer, autoStart, playbackElement, mediaSources, defaultStyleOpts)
+          callback(subtitlesEnabled)
+        })
+        .catch(() => {
+          Plugins.interface.onSubtitlesDynamicLoadError()
+        })
+    } else {
+      new Promise((resolve) => resolve()).then(() => {
         callback(subtitlesEnabled)
```
Luke-Chatburn, line 22: Do we want to return subtitlesEnabled here or false? can't work out what it should be.

ShiningTrapez: I think you can toggle subtitles in the API regardless of if they're availible, so this would make the state be out of sync with it internally

Luke-Chatburn:
Did you look at changing it higher up? https://github.com/bbc/tv-client/blob/master/packages/core-playback/src/subtitles/subtitlesenabled.js#L6

src/subtitles/subtitles.js
```diff
@@ -8,14 +8,20 @@ function Subtitles(mediaPlayer, autoStart, playbackElement, defaultStyleOpts, me
...
+      import("./legacysubtitles.js")
+        .then(({ default: LegacySubtitles }) => {
+          subtitlesContainer = LegacySubtitles(mediaPlayer, autoStart, playbackElement, mediaSources, defaultStyleOpts)
+          callback(subtitlesEnabled)
+        })
+        .catch(() => {
+          Plugins.interface.onSubtitlesDynamicLoadError()
+        })
+    } else {
+      new Promise((resolve) => resolve()).then(() => {
```
eirikbjornr, line 28: Could you add a comment to explain why this is necessary? Maybe link https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#using_microtasks

Could you simplify with `Promise.resolve(() => callback())` or `Promise.resolve().then(callback)`?

eirikbjornr (changes requested):
Gnarly we have to do this – Added some comments. What are thoughts on creating a ticket to revisit this later and make it truly promisified?

src/subtitles/subtitles.js
ShiningTrapez: eb003ed

Luke-Chatburn (approved):

eirikbjornr (approved):

PR #288, opened 2023-04-18
Remove Hybrid Strategy

ShiningTrapez:
📺 What
Remove Hybrid Strategy

🛠 How
Backspace, mostly


.eslintrc.yaml
```diff
@@ -36,6 +36,7 @@ rules:
   unicorn/no-array-reduce: off
   unicorn/prevent-abbreviations: off
   unicorn/catch-error-name: off
+  no-param-reassign: off
```
eirikbjornr, line 39: What's the reason for turning off the param reassign rule and conditional expect?

src/playercomponent.js
```diff
@@ -162,18 +161,26 @@ function PlayerComponent(
...
         break
-      case MediaState.ENDED:
+      }
+      case MediaState.ENDED: {
         onEnded()
         break
+      }
+
+      default: {
+        break
```
eirikbjornr, line 182: What are thoughts on adding some kind of error/telemetry for an unrecognised media state? I'm not suggesting it be part of this ticket. Moreso one to raise as a new ticket?

.eslintrc.yaml
ShiningTrapez: I didn't want to deal with it just trying to get this built, but also param reassign is fairly fundamental to our teardown, since we make everything undefined, and the conditional expect rule is just annoying, it yells at you if you expect inside a promise, which is a really common pattern

eirikbjornr (dismissed):
Core change LGTM! I've raised one suspected bug, and one discussion round a follow on ticket. I'm also interested to hear the reasoning for turning off some linting rules.

.eslintrc.yaml
ShiningTrapez: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md

> The following patterns are warnings:
> ...
>  await foo().catch(error => expect(error).toBeInstanceOf(error));

eirikbjornr: What's the reason to reassign the params to in teardown? Is that guarding for subsequent calls to methods on the `PlayerComponent`? Given we only do it for one I'd argue for just reassigning the param to a local variable in this case.

ShiningTrapez: It does have the "WIP" label which blocks merges

eirikbjornr: Oh! Sorry assumed it was ready – didn't see the WIP label. Second time I review an unfinished PR cus I spot it in `announce-bsp`😅.

ShiningTrapez: It is kinda my fault for not making it a draft, DW, and you should definitely be questioning this if it was otherwise being shipped :D

eirikbjornr (approved):
LGTM🌞

PR #292, opened 2023-04-25
Segmented subs should only be available for seekable devices

matt-stephenson:
📺 What

Only report segmented subtitles as available when the bigscreen-player is configured with a `seekable` `liveSupport`. I.e. segmented subtitles are not available for `playable`, `restartable` or `none`.

🛠 How

Check for the `liveSupport` property on the `window.bigscreenPlayer` config object.

### DIFF
--- a/src/subtitles/subtitles.js
+++ b/src/subtitles/subtitles.js
@@ -3,6 +3,9 @@ import findSegmentTemplate from "../utils/findtemplate"
 
 function Subtitles(mediaPlayer, autoStart, playbackElement, defaultStyleOpts, mediaSources, callback) {
   const useLegacySubs = window.bigscreenPlayer?.overrides?.legacySubtitles ?? false
+  const isSeekableLiveSupport = window.bigscreenPlayer.liveSupport
+    ? window.bigscreenPlayer.liveSupport === "seekable"
+    : true
 
   let subtitlesEnabled = autoStart
   let subtitlesContainer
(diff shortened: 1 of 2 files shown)


Luke-Chatburn (approved):

src/subtitles/subtitles.js
```diff
@@ -3,6 +3,9 @@ import findSegmentTemplate from "../utils/findtemplate"
 
 function Subtitles(mediaPlayer, autoStart, playbackElement, defaultStyleOpts, mediaSources, callback) {
   const useLegacySubs = window.bigscreenPlayer?.overrides?.legacySubtitles ?? false
+  const isSeekableLiveSupport = window.bigscreenPlayer.liveSupport
```
eirikbjornr, line 6: I had trouble following this – Could we separate it into two assertions like `is liveSupport nullish or equal to "seekable"`?

eirikbjornr (changes requested):
LGTM – good testing! I had trouble following the boolean statement, suggested a change.

eirikbjornr (approved):

src/subtitles/subtitles.js
```diff
@@ -69,7 +71,7 @@ function Subtitles(mediaPlayer, autoStart, playbackElement, defaultStyleOpts, me
 
     const isWhole = findSegmentTemplate(url) == null
 
-    return isWhole || !useLegacySubs
+    return isWhole || (!useLegacySubs && isSeekableLiveSupport)
```
Luke-Chatburn, line 74: !isSeekableLiveSupport, don't copy and paste

PR #299, opened 2023-06-16
BADGERS-177 Use `UTCTiming` for wall-clock time

eirikbjornr:
📺 What

Get wall-clock time from the `<UTCTiming>` element present in dynamic DASH manifests.

> Ticket: [BADGERS-177](https://jira.dev.bbc.co.uk/browse/BADGERS-177)

If `serverDate` is passed into bigscreen-player on `init` that timing will be used to derive times from the _first_ manifest only. If a new manifest is fetched, or the current manifest is refreshed, then wall-clock time will be derived from `<UTCTiming>`.

Note:
- This is a PR to merge into branch `badgers-177`, not `master`. I've chunked the ticket into two PRs to make reviews easier.
- Therefore, this ticket does not fix:
  + `timeCorrectionSeconds` not being updated on failover/refresh (closure bug).
  + MPD time being passed as availability time on a simulcast failover by using the POSIX time anchor, causing restart point to be at least 1 minute off.

🛠 How

Fetches and parses wall-clock time from the timing resource defined by the `<UTCTiming>` element on the manifest.

Along the way:
- Promisified `mediasources` to fix tests. Fixed a bug that caused the manifest to be loaded and parsed twice if it failed over when `loadManifest` was called.
- Fixed the `DebugTool` so it formats errors correctly
- Added `warn` method to `DebugTool` to log a deprecation notice.

### DIFF
--- a/docs/tutorials/00-getting-started.md
+++ b/docs/tutorials/00-getting-started.md
@@ -1,17 +1,40 @@
-# Getting Started
-
 ## Installation
 
 ```bash
 npm install bigscreen-player --save
 ```
 
-## Configuration
+## Get To Playback
 
-Bigscreen Player has some global configuration that is needed before initialisation. A *playback strategy* must be configured:
+For a start, let's play back an on-demand video.
+
+To start a playback session you must minimally provide:
+
+- Some configuration. See [Configuration](#configuration)
+- A streaming manifest. See [Minimal Data](#minimal-data)
+- A HTML element to render playback. See [Initialisation](#initialisation)
+
+### Configuration
+
+You must provide a *playback strategy* to use BigscreenPlayer:
 
 ```javascript
-window.bigscreenPlayer.playbackStrategy = 'msestrategy' // OR 'nativestrategy' OR 'hybridstrategy' OR 'basicstrategy'
+window.bigscreenPlayer.playbackStrategy = 'msestrategy' // OR 'nativestrategy' OR 'basicstrategy'
+```
+
+The MSEStrategy uses DASH. It is most likely what you want. More detail in the [documentation on playback strategies](<https://bbc.github.io/bigscreen-player/api/tutorial-playback-strategies.html>).
+
+### Minimal Data
+
+You must provide a manifest and its MIME type.
+
+```javascript
+const minimalData = {
+  media: {
+    type: 'application/dash+xml',
+    urls: [{ url: 'https://example.com/video.mpd' }]
+  }
+}
 ```
 
 ## Initialisation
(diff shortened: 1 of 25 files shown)


docs/tutorials/00-getting-started.md
```diff
@@ -1,17 +1,40 @@
...
 
-Bigscreen Player has some global configuration that is needed before initialisation. A *playback strategy* must be configured:
+You must provide a *playback strategy* to use BigscreenPlayer:
 
 ```javascript
-window.bigscreenPlayer.playbackStrategy = 'msestrategy' // OR 'nativestrategy' OR 'hybridstrategy' OR 'basicstrategy'
+window.bigscreenPlayer.playbackStrategy = 'msestrategy' // OR 'nativestrategy' OR 'basicstrategy'
+```
+
+The MSEStrategy uses DASH. It is most likely what you want. More detail in the [documentation on playback strategies](<https://bbc.github.io/bigscreen-player/api/tutorial-playback-strategies.html>).
```
ShiningTrapez, line 25: It's weird I know but the URL is https://bbc.github.io/bigscreen-player/api/tutorial-Playback%20Strategies.html, with a space in it 

```diff
- The MSEStrategy uses DASH. It is most likely what you want. More detail in the [documentation on playback strategies](<https://bbc.github.io/bigscreen-player/api/tutorial-playback-strategies.html>).
+ The MSEStrategy uses DASH. It is most likely what you want. More detail in the [documentation on playback strategies](<https://bbc.github.io/bigscreen-player/api/tutorial-Playback%20Strategies.html>).
```

ShiningTrapez (approved):
Looks good! Just a comment on documentation, also can't help but note that the Promises in `src/manifest/manifestloader.js` are very similar, could these be collapsed into a function to be more DRY?

docs/tutorials/00-getting-started.md
eirikbjornr: Not since I renamed that file 😛

ShiningTrapez: My bad :sweat_smile:

eirikbjornr:
> [...] can't help but note that the Promises in `src/manifest/manifestloader.js` are very similar, could these be collapsed into a function to be more DRY?

I did think about that but was unsure whether to port our LoadUrlAsync wrapper to another place. But tbf I can introduce a simple wrapper in the manifestloader. TV module network rewrite when 😭

eirikbjornr:
☝️ On this I looked over again and not worth collapsing IMO. A wrapper wouldn't help in terms of performance or readability because each function interacts with LoadUrl slightly differently. That variation would end up in the promise chain instead.

PR #300, opened 2023-06-19
Badgers 177

eirikbjornr:
📺 What

Combines PR #299 and #302 to:

- add a feature: Get wall-clock time from the `<UTCTiming>` element present in dynamic DASH manifests,
- and a fix: Pass current MPD time to Dash.js to restart playback with a new source at the same point in time on CDN failover.

> Ticket: [BADGERS-177](https://jira.dev.bbc.co.uk/browse/BADGERS-177)

🛠 How

See details of each child branch. TL;DR is:

- Fetches and parses wall-clock time from the timing resource defined by the `<UTCTiming>` element on the manifest if a `serverDate` is not provided. If a `serverDate` is provided, `<UTCTiming>` is used for the second manifest load onwards.
- Modify MSEStrategy to pass failover time as an MPD time instead of a video time.


src/bigscreenplayer.test.js
```diff
@@ -786,14 +786,12 @@ describe("Bigscreen Player", () => {
       initialiseBigscreenPlayer({ windowType: WindowTypes.SLIDING })
 
       const callback = jest.fn()
-      const endOfStreamWindow = bigscreenPlayerData.time.windowEndTime - 2
+      const endOfStreamWindow = 100 - 2
```
ShiningTrapez, line 789: Can we just make this 98? :D

ShiningTrapez, line 789: We use `100` in a few locations in this file, can we make it a const, so if it changes for whatever reason in the future, you don't need to update it here, L793, and L820

src/debugger/debugtool.test.js
```diff
@@ -1,55 +1,113 @@
...
+    })
   })
 
-  it("overwrites a keyvalue entry to the chronicle if that keyvalue already exists", () => {
-    const testObj = { key: "akey", value: "something" }
-    const testObj1 = { key: "bitrate", value: "1000" }
-    const testObj2 = { key: "bitrate", value: "1001" }
+  describe("logging a warning", () => {
+    it("takes a string", () => {
+      DebugTool.warn("you're using a deprecated thingie!")
```
ShiningTrapez, line 57: Maybe we should const this string too so we don't need to update it in multiple locations again, the test on L67 can interpolate it in?

src/manifest/manifestloader.js
```diff
@@ -88,13 +90,15 @@ function getStreamUrl(data) {
...
     if (/\.mpd($|\?.*$)/.test(mediaUrl)) {
-      retrieveDashManifest(mediaUrl, { windowType, initialWallclockTime, onError, onSuccess })
-    } else if (/\.m3u8($|\?.*$)/.test(mediaUrl)) {
-      retrieveHLSManifest(mediaUrl, { windowType, initialWallclockTime, onError, onSuccess })
-    } else {
-      onError("Invalid media url")
+      return retrieveDashManifest(mediaUrl, { windowType, initialWallclockTime })
+    }
+
+    if (/\.m3u8($|\?.*$)/.test(mediaUrl)) {
```
ShiningTrapez, line 98: I know you didn't write this regex, but we could just make the query param an optional match like `/\.m3u8(\?.*)?$/`, I know regex comprehension isn't most people's strong suite in the first place, but I think this reads better since you can interpret it entirely forward (As a bonus, it's ever so slightly faster in the case where there's a query param too, taking one less step).

ShiningTrapez (approved):
LGTM, just got some minor style things I won't block on

src/bigscreenplayer.test.js
eirikbjornr: IMO that's optimising for the current state of the tests and may make it harder to change in the future. I tend to make my tests as unoptimised as possible because find/replace can do the job

src/debugger/debugtool.test.js
eirikbjornr: Same as previous comment

PR #301, opened 2023-06-21
Fix live delay clamping for OD content

matt-stephenson:
📺 What

Fixes issue where the `liveDelay` is subtracted from the end even for On Demand content.

🛠 How

No longer subtract this value unless using the `getClampedTimeForLive` function.


ShiningTrapez:
Superseded by #351

PR #303, opened 2023-06-28
[spike] Use signal-based triggers

eirikbjornr:
📺 What

Inject a signal object to use event-based triggers.

🛠 How

Create a DebugController with a signal (EventTarget). Trigger events in BSP through the controller.

👀 See

Video of using the debug controller to trigger a failover and manifest reload: https://www.dropbox.com/s/8xdqbng5s0j97qp/badger-spike-signal-triggers.mov?dl=0

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,14 +7,34 @@
     </head>
     <body>
         <script type="module">
-            import { BigscreenPlayer } from "./dist-local/esm/main.js"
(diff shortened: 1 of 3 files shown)


markhouldridge (commented):
Looks interesting and found this while reviewing ticket, BADGER-77.  Would be good to check if this is still something desired to improve debugging.  I could see this being useful, around forcing CDN failures from EPT, as an action button.

PR #310, opened 2023-09-21
BADGERS-107 Prevent seek to end

jamesbungay-bbc:
📺 What

Prevents seeking to the end of media to fix issues observed on some devices where seeking to/past the duration does not cause an ended event to be thrown.

🛠 How

Added `SEEK_END_PADDING` of 0.1 (seconds) to `msestrategy.js`, which is used to clamp the seekable range to the greatest of itself and the pre-existing LiveDelay.


ShiningTrapez (approved):
LGTM

ShiningTrapez (approved):

PR #311, opened 2023-10-12
Add guard around _stateUpdateCallback

matt-stephenson:
📺 What

Prevents an unhandled exception when the player is torn down, but an async callback from a CDN failover error returns.

🛠 How

As the player has been torn down, we can safely ignore the return and just guard against the _stateUpdateCallback existing.

This PR also updates a couple of unit tests that were found to not be implemented correctly whilst investigating this unhandled exception.

### DIFF
--- a/src/playercomponent.test.js
+++ b/src/playercomponent.test.js
@@ -1050,6 +1050,7 @@ describe("Player Component", () => {
       jest.useFakeTimers()
 
       setUpPlayerComponent()
+      forceMediaSourcesError = true
 
       return StrategyPicker.default().then(() => {
         mockStrategy.mockingHooks.fireEvent(MediaState.WAITING)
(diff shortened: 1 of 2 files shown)


src/playercomponent.test.js
```diff
@@ -1058,7 +1059,9 @@ describe("Player Component", () => {
 
         jest.advanceTimersByTime(30000)
 
-        expect(mockStateUpdateCallback.mock.calls[0][0].data.state).not.toBe(MediaState.FATAL_ERROR)
+        // expect 1 call as player goes into WAITING when event is fired above, but should not
+        // have a call after the time advances as the timer will have been cleared
+        expect(mockStateUpdateCallback.mock.calls).toHaveLength(1)
```
eirikbjornr, line 1064:
```suggestion
        expect(mockStateUpdateCallback).toHaveBeenCalledTimes(1)
```

src/playercomponent.test.js
```diff
@@ -1079,7 +1083,9 @@ describe("Player Component", () => {
 
         jest.advanceTimersByTime(5000)
 
-        expect(mockStateUpdateCallback.mock.calls[0][0].data.state).not.toBe(MediaState.FATAL_ERROR)
+        // expect 1 call as player goes into WAITING when fireError is called above, but should not
+        // have a call after the time advances as the timer will have been cleared
+        expect(mockStateUpdateCallback.mock.calls).toHaveLength(1)
```
eirikbjornr, line 1088:
```suggestion
        expect(mockStateUpdateCallback).toHaveBeenCalledTimes(1)
```

eirikbjornr (changes requested):
LGTM! Requested some changes to use specific assertion functions

eirikbjornr (approved):

PR #326, opened 2024-03-13
BADGERS-380 feat(chronicle): tweak the data

eirikbjornr:
📺 What

Tweak data out of Chronicle

🛠 How

Add apicall trace
Add playback rate metric
Log seamless failover errors
Stop sending stalled events
Take current element time from the media element

### DIFF
--- a/src/playbackstrategy/basicstrategy.js
+++ b/src/playbackstrategy/basicstrategy.js
@@ -1,3 +1,4 @@
+import DebugTool from "../debugger/debugtool"
 import MediaState from "../models/mediastate"
 import WindowTypes from "../models/windowtypes"
 import MediaKinds from "../models/mediakinds"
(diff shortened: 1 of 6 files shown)


src/playbackstrategy/basicstrategy.js
```diff
@@ -36,14 +37,16 @@ function BasicStrategy(mediaSources, windowType, mediaKind, playbackElement) {
...
-      mediaElement.src = mediaSources.currentSource()
-      setStartTime(startTime)
-      mediaElement.load()
+
+      return
     }
+
+    mediaElement.src = mediaSources.currentSource()
+    setStartTime(startTime)
+    mediaElement.load()
```
ShiningTrapez, line 49: If we return in the else, can't lines 47 to 49 be inside the if, and then if that's true, can't we just eliminate the else entirely?

ShiningTrapez (changes requested):

ShiningTrapez (approved):

PR #327, opened 2024-03-14
BADGERS Fix debugview

eirikbjornr:
📺 What

Title.

🛠 How

Add missing trace types to the debug view
Stop updating time from bigscreenplayer in favour of writing time per strategy

### DIFF
--- a/src/debugger/debugviewcontroller.ts
+++ b/src/debugger/debugviewcontroller.ts
@@ -151,7 +151,7 @@ class DebugViewController {
   }
 
   private serialiseDynamicEntry(entry: DynamicEntry): string {
-    let formattedData: string
+    let formattedData: string | undefined
 
     const { category } = entry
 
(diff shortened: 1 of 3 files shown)


ShiningTrapez (approved):

src/debugger/debugviewcontroller.ts
```diff
@@ -167,15 +167,12 @@ class DebugViewController {
...
-
-      default:
-        throw new TypeError(`Unrecognised Entry Category: ${category}`)
     }
 
     const sessionTime = new Date(entry.sessionTime)
     const formatedSessionTime = `${formatDate(sessionTime)}.${zeroPadMs(sessionTime.getUTCMilliseconds())}`
 
-    return `${formatedSessionTime} - ${formattedData}`
+    return `${formatedSessionTime} - ${formattedData satisfies string}`
```
matt-stephenson, line 175: Should this be `${formattedData} satisfies string` ?

matt-stephenson (changes requested):
Looks good, just one question

src/debugger/debugviewcontroller.ts
ShiningTrapez: No, `[satisfies](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) string` is part of the code

PR #336, opened 2024-05-20
Defer plugin exceptions

Luke-Chatburn:
📺 What

Defer plugin exceptions.

https://jira.dev.bbc.co.uk/browse/BADGERS-309

🛠 How

Use callCallbacks in the plugins.

### DIFF
--- a/src/plugins.test.js
+++ b/src/plugins.test.js
@@ -1,10 +1,6 @@
 import plugins from "./plugins"
 
 describe("Plugins", () => {
-  beforeEach(() => {
-    // mockReset();
-  })
-
   it("Calls a registered plugin on interface invocation", () => {
     const fatalErrorPlugin = {
       onFatalError: jest.fn(),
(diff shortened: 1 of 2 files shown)


src/plugins.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+    jest.useFakeTimers()
+
+    try {
+      expect(() => {
+        plugins.interface.onFatalError()
+      }).not.toThrow()
+      jest.advanceTimersByTime(1)
+    } catch (error) {
+      // eslint-disable-next-line jest/no-conditional-expect
```
ShiningTrapez, line 41: I feel this could be implemented by adding the validation logic inside the plugin instead of catching an error and then ESLint Disabling a bunch of checks

ShiningTrapez (changes requested):

ShiningTrapez (approved):

PR #337, opened 2024-05-29
Do not use string.split on non-segmented subtitles

edmorrish:
📺 What

Due to a performance issue running string.split on large subtitle segments, replace the use of string.split with a more efficient replacement.

🛠 How

Replace string.split with alternative use of string.replace.

### DIFF
--- a/src/subtitles/imscsubtitles.js
+++ b/src/subtitles/imscsubtitles.js
@@ -83,7 +83,9 @@ function IMSCSubtitles(mediaPlayer, autoStart, parentElement, mediaSources, defa
         }
 
         try {
-          const xmlText = isSubtitlesWhole() ? responseText : responseText.split(/<\?xml[^?]+\?>/i)[1] || responseText
+          const xmlText = isSubtitlesWhole()
+            ? responseText.replace(/^.*<\?xml[^?]+\?>/i, "")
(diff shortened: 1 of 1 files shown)


src/subtitles/imscsubtitles.js
```diff
@@ -83,7 +83,10 @@ function IMSCSubtitles(mediaPlayer, autoStart, parentElement, mediaSources, defa
         }
 
         try {
-          const xml = fromXML(responseText.split(/<\?xml[^?]+\?>/i)[1] || responseText)
+          const xmlText = isSubtitlesWhole()
+            ? responseText.replace(/^.*<\?xml[^?]+\?>/i, "")
+            : responseText.split(/<\?xml[^?]+\?>/i)[1] || responseText
```
eirikbjornr, line 88: What's the reason we split for live? Do you reckon it'll reduce test effort?

edmorrish: No idea, intention for this is to reduce the blast radius of the change and limit it to VOD. There isn't really a performance issue with running split like this on segmented streams because they're only short.

eirikbjornr: Yea fair – one for us to look into the performance of this further

PR #343, opened 2024-07-23
Fix Pause before Seek Scenario for Sliding Window

ShiningTrapez:
📺 What
Fixes an issue where a user pause in a session preceding a seek in Sliding Window content could erroneously offset the seek by the time paused.

🛠 How
Distinguish between a `USER` and `APP` pause in calculating offset time

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -582,13 +582,12 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
     }
 
     if (windowType === WindowTypes.SLIDING) {
-      console.log(`Sliding Window Paused Time: ${new Date(slidingWindowPausedTime).toString()}`)
       const dvrInfo = mediaPlayer.getDashMetrics().getCurrentDVRInfo(mediaKind)
       const offset = TimeUtils.calculateSlidingWindowSeekOffset(
         time,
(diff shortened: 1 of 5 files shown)


src/playbackstrategy/msestrategy.js
```diff
@@ -700,7 +701,13 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
         startAutoResumeTimeout()
       }
     },
-    play: () => mediaPlayer.play(),
+    play: () => {
+      if (windowType === WindowTypes.SLIDING) {
+        slidingWindowPausedTime = 0
```
Luke-Chatburn, line 706: Is this needed?

If you only ever set it on an app pause then this will always result in a seek and be reset.

ShiningTrapez: This is just an artifact of an earlier attempt but I feel like it doesn't hurt to keep it in just to make sure the state is reset

Luke-Chatburn: Whilst I agree in this instance it isn't that big a deal but don't feel it's good to keep around dead/redundant code.

ShiningTrapez: I'll remove it

Luke-Chatburn (approved):

PR #348, opened 2024-09-11
BADGERS-350: upgrade Actions workflows to use node20 & latest action versions

tom-coward:
📺 What

Upgrades all Actions workflows to use Node 20 (upgrading from 16), and all actions to their latest versions.

🛠 How

Define `node-version` as 20 in `setup-node` steps.
Define versions of all actions to the current latest.


.github/workflows/codeql-analysis.yml
```diff
@@ -39,11 +39,11 @@ jobs:
...
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v1
+        uses: github/codeql-action/init@v3
         with:
           languages: ${{ matrix.language }}
```
ShiningTrapez, line 48: You could inline this Matrix in the same way you did for the Node Versions, just to be consistent

ShiningTrapez (dismissed):
LGTM!

.github/workflows/codeql-analysis.yml
tom-coward: Good spot - have done this

ShiningTrapez (dismissed):

ShiningTrapez (approved):

tom-coward:
Bypassing checks to get this merged as codeql can't find an old codeql analysis run associated with the PR to compare to after switching out the matrix languages to be hardcoded - we think this should be fine for future PRs but will verify this after merging:
image
image

PR #349, opened 2024-09-12
BADGERS-350: revert to using language matrix in codeql

tom-coward:
📺 What

Reverts to using a matrix to define which languages to run CodeQL analysis on.

🛠 How

Reintroduce the `languages` strategy matrix, and reference it when defining `languages` of the CodeQL init action.


tom-coward:
Had to manually delete (via. GH API) the previous codeql analysis run on master with no environment (from #348's merge commit) and codeql is now happy again after reverting to using the matrix, as it isn't trying to compare runs with references that don't match up on master & the PRs (see https://github.com/github/codeql-action/issues/1179)

ShiningTrapez (approved):

PR #351, opened 2024-09-17
Don't apply `liveDelay` to `WindowTypes.STATIC`

ShiningTrapez:
📺 What
Don't apply `liveDelay` to `WindowTypes.STATIC`, supersedes #301 

🛠 How
Add a checks for `windowType` in `getClampedTime()` and use the implementation from `getClampedTimeForLive()` for `WindowTypes.SLIDING`, the original implementation for `WindowTypes.GROWING`, and #301, with `seekDurationPadding`, for `WindowTypes.STATIC`. Refactored into a single expression with ternary checks preceeding. 

👀 See
#301

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -454,15 +454,12 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
   }
 
   function getClampedTime(time, range) {
-    if (windowType === WindowTypes.SLIDING) {
-      return Math.min(Math.max(time, 0), mediaPlayer.getDVRWindowSize() - Math.max(liveDelay, seekDurationPadding))
-    }
+    const rangeStart = windowType === WindowTypes.SLIDING ? 0 : range.start
+    const rangeEnd = windowType === WindowTypes.SLIDING ? mediaPlayer.getDVRWindowSize() : range.end
+    const correction =
+      windowType === WindowTypes.STATIC ? seekDurationPadding : Math.max(liveDelay, seekDurationPadding)
 
-    if (windowType === WindowTypes.GROWING) {
-      return Math.min(Math.max(time, range.start), range.end - Math.max(liveDelay, seekDurationPadding))
-    }
-
-    return Math.min(Math.max(time, range.start), range.end - seekDurationPadding)
+    return Math.min(Math.max(time, rangeStart), rangeEnd - correction)
   }
 
   function load(mimeType, playbackTime) {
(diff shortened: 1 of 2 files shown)


src/playbackstrategy/msestrategy.js
```diff
@@ -454,7 +454,15 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
   }
 
   function getClampedTime(time, range) {
-    return Math.min(Math.max(time, range.start), range.end - Math.max(liveDelay, seekDurationPadding))
+    const vod = windowType === WindowTypes.STATIC
```
Luke-Chatburn, line 457: can we change that to static? this will apply for audio/video

ShiningTrapez: I don't understand the question :face_with_spiral_eyes: I think this is independent of Audio/Video yes, but I don't do anything specific to audio/video?

src/playbackstrategy/msestrategy.js
```diff
@@ -586,23 +594,20 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
...
+      dvrInfo.range.start,
+      timeCorrection,
+      slidingWindowPausedTime
+    )
+    slidingWindowPausedTime = 0
 
-      return getClampedTimeForLive(offset)
-    }
-    return getClampedTime(time, getSeekableRange())
+    return getClampedTime(offset)
```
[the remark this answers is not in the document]
ShiningTrapez: It's not an object? https://github.com/bbc/bigscreen-player/blob/48bc4db62dc4aa591e8da3d374a52f4729de99bc/src/utils/timeutils.ts#L24-L37

src/playbackstrategy/msestrategy.js
matt-stephenson: The const is called `vod` - but shouldn't really be names as something video specific, as it might be audio content.

src/playbackstrategy/msestrategy.js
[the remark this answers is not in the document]
matt-stephenson: Yeah, you're right, my bad! I wonder if you could pass the range here for sliding as the second argument, then the arguments are consistent, and you wouldn't need the ternaries in `getClampedTime`? (although would need to check the file for other calls to it)

src/playbackstrategy/msestrategy.js
ShiningTrapez: Ah yes, I can change that, I think that's just a factoring leftover

src/playbackstrategy/msestrategy.js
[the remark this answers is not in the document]
ShiningTrapez: I did it this way to centralise the logic, I didn't like that there was a one time use `getClampedTimeForLive` miles away from `getClampedTime` and that's essentially recreating it

matt-stephenson (approved):

Issue #95, opened 2019-09-16
`addEventCallback` inconsistent across playback strategies


tsadler1988:
As of #90, `msestrategy` supports multiple callbacks, whereas other playback strategies do not.

PR #352, opened 2024-09-27
Add support for Dash.js rendered subtitles

ryanmccartney:
📺 What

Extends subtitle functionality by adding the ability to use Dash.js rendered subtitles if declared in an MPD rather than using the existing side-chain.

Allows the rendering of subtitles delivered in fragmented chunks, as used for low-latency playback.

It does not implement the current customisation options that are available in the existing side-chain as Dash.js provides no interface to access the customisation options present in the BBC fork of imscJS.

A currently open PR [imscJS #257](https://github.com/sandflow/imscJS/pull/257) would enable customisation on upstream imscJS and work on [Dash.js](https://github.com/bbc/rd-dash.js/tree/imsc-styling) [(imscJS](https://github.com/bbc/rd-dash.js/tree/imsc-styling)[ Styling](https://github.com/bbc/rd-dash.js/tree/imsc-styling) would allow this functionality to be added retrospectively.

🛠 How
By setting a new override - `dashSubtitles`, the side-chain is replaced by a new `dashsubtitles` object which enables text tracks in the MSE player

✅ Testing [Semi-optional]

Tests have been added to cover changes to the subtitle selector. In addition, coverage of the new dashsubtitles.js object is provided.

♿ Accessibility [optional]

Provides optional subtitles when using low-latency playback through Bigscreen player without modifications to the existing side-chain rendering.


src/playbackstrategy/msestrategy.js
```diff
@@ -548,6 +554,21 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
     mediaPlayer.on(DashJSEvents.GAP_JUMP, onGapJump)
     mediaPlayer.on(DashJSEvents.GAP_JUMP_TO_END, onGapJump)
     mediaPlayer.on(DashJSEvents.QUOTA_EXCEEDED, onQuotaExceeded)
+    mediaPlayer.on(DashJSEvents.TEXT_TRACKS_ADDED, disableTextTracks)
+  }
+
+  function disableTextTracks() {
```
eirikbjornr, line 560: What calls this function? Should `setSubtitles` call this when `state === false`?

eirikbjornr (commented):
LGTM overall! Got one question on `setSubtitles` and `disableTextTracks`

src/playbackstrategy/msestrategy.js
ryanmccartney: Hi @eirikbjornr thanks for taking a look. The thought behind this is that calling `disableTextTracks` when a `TECT_TRACK` is added is a failsafe way of preventing Dash.js from ever showing subtitles before BSP has signalled a need for them. As a result, I don't believe it's required on `setSubtitles(false)`, if it's helpful to add it in for consistency, I'm happy to do so.

matt-stephenson:
Just a thought on naming. I wonder if `embeddedSubtitles` might be a little clearer. Although it does work, as subtitles are in the dash manifest, the current side-loading of subs still happens during DASH playback.

ryanmccartney:
> Just a thought on naming. I wonder if `embeddedSubtitles` might be a little clearer. Although it does work, as subtitles are in the dash manifest, the current side-loading of subs still happens during DASH playback.

That makes sense, I'll adjust the naming to reflect this

src/playbackstrategy/msestrategy.js
eirikbjornr: Ah I didn't spot it was added as a handler on `TEXT_TRACKS_ADDED` – that all makes sense now

ryanmccartney:
References to Dash.js subtitles have now been updated to Embedded subtitles

ryanmccartney:
I've submitted a PR to `bbc/dash.js` [#95](https://github.com/bbc/dash.js/pull/95) to use `bbc/imsc.js` in `bbc/dash.js` and expose the customisation options through the `dash.js` settings interface. This PR could be further expanded to provide subtitle customisation to match the current offering via the side-chain. Alternatively, this work could be done in a seperate PR.

ryanmccartney:
PR now brought up to date with 9.0.2

ryanmccartney:
Replace by [#383](https://github.com/bbc/bigscreen-player/pull/383)

PR #361, opened 2024-10-30
Add target-latency and current-latency to debug view

ryanmccartney:
📺 What

Adds MSE target latency and current latency reporting to the debug menu. Useful when debugging low-latency content, particularly on TV hardware where some decoder behaviours do not match what might be expected

🛠 How

Uses existing `DebugTool.dynamicMetric()` method to add these metrics each time a `timeupdate` event is called.

✅ Testing

Added underlying dash.js functions to the `mockDashInstance`.  No additional unit tests were added.

👀 See

New fields are seen below when the MSE strategy is chosen. The values` NaN` and 0 are seen when VOD content is played.

Screenshot 2024-10-30 at 17 06 22

The same changes are shown again, but using live low-latency content. Extra fields, along with the existing playback rate field, can be used to determine whether low-latency content is performing as expected.

Screenshot 2024-10-30 at 17 06 54

♿ Accessibility [optional]

[Any accessibility features or considerations that this PR addresses should be listed here]

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -188,6 +188,8 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
 
   function onTimeUpdate() {
     DebugTool.updateElementTime(mediaElement.currentTime)
+    DebugTool.dynamicMetric("current-latency", mediaPlayer.getCurrentLiveLatency())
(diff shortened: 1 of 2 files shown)


ShiningTrapez (approved):

ShiningTrapez (changes requested):
Needs Chronicle Changes - Working on it, just blocking merge

PR #362, opened 2024-12-09
Add onPlaybackRateChanged and latency to onPlayerInfoUpdated

ryanmccartney:
📺 What

Exposes PLAYBACK_RATE_CHANGED event from Dash.js to the plugin interface - part of larger work to support low-latency playback. This change allows information on the current playback rate to be captured.

Additionally, it also adds `onDashMetrics` to the plugin interface to expose more detailed player information to the plugin. Specifically, this would be used to determine the current latency as follows;

```
const onDashMetrics = (event) => {
        const dvrInfo = event.getCurrentDVRInfo();
        const latency = dvrInfo.range.end - dvrInfo.time
};
```
These are important for identifying any devices which have problems with low-latency playback.

E.g.) If logs suggest the playback rate is greater than 1.0 but the latency reported does not reduce we can reasonably assume that the device is struggling to playback at the speed requested by the player.

🛠 How

Adds an additional event `PLAYBACK_RATE_CHANGED` to `DashJSEvents` in the `MSEStrategy`. On this event being triggered it is passed to a newly added plugin function `Plugins.interface.onPlaybackRateChanged`.

Also adds a plugin interface `Plugins.interface.onDashMetrics` to an existing event `METRIC_ADDED` to provide more detailed access to metrics with plugins. This could potentially be more tightly scoped to `onDvrInfoChanged`.

✅ Testing

No new tested added for this change, happy to add any ones deemed necessary on review.


ryanmccartney:
Closing without merging as this functionality has been provided by another PR

PR #365, opened 2025-01-29
BADGERS-575: [BREAKING] Audio Described Unification

ShiningTrapez:
📺 What
Support a "Source Based" method of providing an, alternative, Audio Described stream with a new Media Block consisting of alternative sources which can be switched to and from mid stream. This approach is primarily intended for Native Strategy Devices but can work on any device given the streams are present and passed correctly (Assuming the device can play the stream and so on).

🛠 How
* Refactor `MediaSources` to keep track of and index into a record of Connection Types, the design has been left open to make further additions along these lines easy in the future.
* Expose a new, Unified, Audio Described API which can seamlessly be used with both this implementation, and the MSE Specific Implementation:
  * `registerForAudioDescribedChanges` (Previously `registerForBroadcastMixADChanges`)
  * `unregisterForAudioDescribedChanges` (Previously `unregisterForBroadcastMixADChanges`)
  * `isAudioDescribedAvailable` (Previously `isBroadcastMixADAvailable`)
  * `isAudioDescribedEnabled` (Previously `isBroadcastMixADEnabled`)
  * `setAudioDescribed` (Previously `setBroadcastMixADEnabled`)

* Adds a new `enableAudioDescribed` option to Bigscreen's Init Options to start up with the AD Stream.
* Various QoL Fixes and Refactors as encountered during development.

:warning: TODO
- [x] Native Playback Strategy Tests
- [x] Test for BSP passing `enableAudioDescribed` through
- [x] `MediaSources` tests
- [x] `PlayerComponent` tests
- [x] Verify Failover Scenarios 
- [x] Verify MSE Implementation
- [x] Write Docs for Generic Method
- [x] Cleanup
  - [x] Consider removing `audio-described-cdns-available` in Chronicle and just add it to `cdns-available` when AD Stream is active? (Testers commented that current setup is confusing)

:rotating_light: BROKEN
* [x] Callback is not fired if AD is on by default


src/mediasources.ts
```diff
@@ -173,10 +183,48 @@ function MediaSources() {
...
     return new Promise((resolve) => resolve(loadManifest()))
   }
 
+  function replace(sources: Connection[]): Promise<void> {
+    const sameObject = (first: object, second: object) => {
+      const normStringify = (obj: object) => JSON.stringify(obj)?.split("").sort().join("")
+      return normStringify(first) === normStringify(second)
+    }
+
+    if (sameObject(sources, mediaSources)) return Promise.resolve()
```
ShiningTrapez, line 196: Do we need to check if the passed in array equals current sources + failover?

src/bigscreenplayer.test.js
```diff
@@ -1533,39 +1532,40 @@ describe("Bigscreen Player", () => {
...
+      expect(mockPlayerComponentInstance.setAudioDescribed).toHaveBeenCalledWith(true)
+      expect(mockPlayerComponentInstance.setAudioDescribed).not.toHaveBeenCalledWith(false)
 
-      bigscreenPlayer.setBroadcastMixADEnabled(false)
+      bigscreenPlayer.setAudioDescribed(false)
 
-      expect(mockPlayerComponentInstance.setBroadcastMixADOn).toHaveBeenCalledTimes(1)
-      expect(mockPlayerComponentInstance.setBroadcastMixADOff).toHaveBeenCalledTimes(1)
+      expect(mockPlayerComponentInstance.setAudioDescribed).toHaveBeenCalledTimes(2)
+      expect(mockPlayerComponentInstance.setAudioDescribed).toHaveBeenCalledWith(true)
```
eirikbjornr, line 1547: You could save a couple lines here with `toHaveBeenNthCalledWith`

src/playbackstrategy/msestrategy.js
```diff
@@ -603,23 +610,37 @@ function MSEStrategy(
 
       // FIX: Dash.js briefly returns `null` on a failover for the first time update
       if (dvrInfo) {
-        return { start: dvrInfo.range.start, end: dvrInfo.range.end - liveDelay }
+        const seekableRange = { start: dvrInfo.range.start, end: dvrInfo.range.end - liveDelay }
+        // Save good seekable range and duration values
+        cached.seekableRange = Utils.clone(seekableRange)
+        cached.duration = getDuration()
```
eirikbjornr, line 616: What's the reason you cache this here instead of in `getDuration`?

src/playbackstrategy/msestrategy.js
```diff
@@ -603,23 +610,37 @@ function MSEStrategy(
...
   }
 
   function getCurrentTime() {
-    return mediaElement?.currentTime ?? 0
+    const currentTime = mediaElement?.currentTime
+    if (currentTime && !isNaN(currentTime)) {
+      cached.currentTime = currentTime
+    }
+
+    return currentTime || cached.currentTime
```
eirikbjornr, line 638: Maybe a silly question – what happens if you seek back to zero? I guess it would update once time progressed to 0.1(?) I don't think that's an issue you need to solve now, feels like we need a better long-term solution than ignoring zeroes. Maybe move ready helper into each strategy? 🤔

src/playbackstrategy/msestrategy.js
```diff
@@ -757,13 +784,15 @@ function MSEStrategy(
 
       // FIX: Dash.js briefly returns `null` on a failover for the first time update
       if (dvrInfo) {
-        const { range } = dvrInfo
+        const seekableRange = { start: dvrInfo.range.start, end: dvrInfo.range.end - seekDurationPadding }
```
eirikbjornr, line 787: What's the reason you're caching seekable range accounting for live delay and seekable range accounting for seek duration padding to the same value? I recommend caching the range out of Dash.js without accounting for either, then apply the live delay or padding in each function. What's thoughts?

src/debugger/chronicle.ts
```diff
@@ -55,6 +55,7 @@ type RepresentationVideo = CreateMetric<"representation-video", [qualityIndex: n
 type SeekableRange = CreateMetric<"seekable-range", [start: number, end: number]>
 type SubtitleCDNsAvailable = CreateMetric<"subtitle-cdns-available", string[]>
 type SubtitleCurrentUrl = CreateMetric<"subtitle-current-url", string>
+type AudioDescribedCDNsAvailable = CreateMetric<"audio-described-cdns-available", string[]>
```
eirikbjornr, line 58: You can get rid of this since you refactored how you do the source switch

src/playbackstrategy/msestrategy.js
```diff
@@ -681,31 +702,37 @@ function MSEStrategy(
...
   }
 
-  function setBroadcastMixADOff() {
+  function setAudioDescribedOff() {
     const audioTracks = mediaPlayer.getTracksFor("audio")
     const mainTrack = audioTracks.find((track) => track.roles.includes("main"))
     mediaPlayer.setCurrentTrack(mainTrack)
+
+    // Feature Parity with Native
+    if (isPaused()) mediaPlayer.play()
```
Luke-Chatburn, line 726: Do we want this change or do we want to leave it up to the consumers?

eirikbjornr (changes requested):
Great work – there's some changes left behind from refactors that I think will save you some work, and think I found an issue in the MSEStrategy caching. Not every comment I've left is a request for change, but you can't really separate those out. Don't hesitate to ask if it's unclear whether something is a request for change

eslint.config.js
```diff
@@ -176,6 +176,9 @@ export default tseslint.config(
       "unicorn/prefer-optional-catch-binding": "off",
       "unicorn/prefer-string-replace-all": "off",
 
+      // Not Sure about this one
+      "unicorn/prefer-set-has": "off",
```
markhouldridge, line 180: What's the reason for changing this?

src/playbackstrategy/msestrategy.js
ShiningTrapez: Probably need a discussion about this, maybe config?

src/playbackstrategy/msestrategy.js
ShiningTrapez: @terminal-overflow wrote this, any idea what your thinking was here?

src/debugger/chronicle.ts
ShiningTrapez: I thought this was gone tbh

eslint.config.js
ShiningTrapez: We can't use Set on TVs, there is a Polyfill for it in `tv-client` but no-one uses it, it's safer to not be the first, especially since we're the app core

src/playbackstrategy/msestrategy.js
ShiningTrapez: I think we can fix that easily by checking explicitly for 0, JS strikes again I guess

src/playbackstrategy/nativestrategy.js
Luke-Chatburn, line 1: Love this functional change, great work

src/playbackstrategy/msestrategy.js
ShiningTrapez: This was done specifically in response to the Product Decision I got from Tom W. wrt the difficulties implementing pause after switch on all devices and the existing behaviour on AD VoD

eirikbjornr: To me it's a question of "do we need parity with native, or can MSE be the 'better' way preserving state". IMO product features should live in the products, and BSP and CorePlayback should enable those features. What's your thoughts?

src/playbackstrategy/msestrategy.js
ShiningTrapez: @terminal-overflow wrote a Unit Test for this specifically, is 0 considered a bad value?

eirikbjornr: My point is `currentTime` is zero both during failover (aka before the media element is configured) and on a seek to start. There's no way around that afaik. `currentTime` is never `undefined` or `null`. As I said above impact is small though afaik

src/playbackstrategy/msestrategy.js
ShiningTrapez: It'll have to be retested if I move this into iPlayer :weary:

src/playbackstrategy/msestrategy.js
ShiningTrapez: Coel and I think we can remove this actually given it was related to the pause after switch behaviour

eslint.config.js
eirikbjornr: Can you move it up to the "Compatibility" block so the reason it's turned off is clearer

src/mediasources.ts
```diff
@@ -19,19 +20,33 @@ type FailoverParams = {
...
-  let subtitlesSources: CaptionsConnection[] = []
-  // Default 5000 can be overridden with media.subtitlesRequestTimeout
+
+  // Can be overridden with media.subtitlesRequestTimeout
   let subtitlesRequestTimeout = 5000
   let failoverResetTimeMs = 120000
-  let failoverSort: ((sources: Connection[]) => Connection[]) | null = null
 
-  function init(media: MediaDescriptor): Promise<void> {
+  function init(media: MediaDescriptor, setAudioDescribedOn?: boolean): Promise<void> {
```
markhouldridge, line 49: Just a preferance, but I'd go with `isAudioDescribedEnabled` instead of `setAudioDescribedOn`

ShiningTrapez: isX is a getter, setX is a setter

src/playercomponent.js
```diff
@@ -88,20 +116,57 @@ function PlayerComponent(
...
+          availabilityTimeInMilliseconds,
+          mediaSources.time().availabilityStartTimeInMilliseconds
+        )
+
+        tearDownMediaElement()
+        loadMedia(mediaMetaData.type, presentationTimeInSeconds)
+      })
+      .catch(() => {
+        bubbleFatalError(false, {
+          code: "0000", // change code?
```
eirikbjornr, line 154: Is this code correct? How is this surfaced in products?

src/playercomponent.js
```diff
@@ -88,20 +116,57 @@ function PlayerComponent(
     return playbackStrategy?.getSeekableRange()
   }
 
-  function isBroadcastMixADAvailable() {
-    return playbackStrategy && playbackStrategy.isBroadcastMixADAvailable?.()
+  function isAudioDescribedAvailable() {
+    const genericAD = mediaSources.isAudioDescribedAvailable()
```
eirikbjornr, line 120: What's thoughts on a less "generic" (pun not intended but now I've typed this it's definitely intended – language is a paradox) name? `sourceBasedAD` is the first thing that comes to my mind

eirikbjornr (changes requested):
Spotted a potential issue: Is the error code surfaced on a failed AD switch correct?

src/playbackstrategy/msestrategy.js
terminal-overflow: I've moved this `.duration` setting in https://github.com/bbc/bigscreen-player/pull/365/commits/78c4a990511db9bae86a3c17d179664af2d8fef5.

src/playercomponent.js
ShiningTrapez: I'm open to changing this, I didn't think much about this name

src/playercomponent.js
ShiningTrapez: Ye, I think we never changed this from Matt's Spike

src/playercomponent.js
terminal-overflow: I like `sourceBased`. This has been addressed in https://github.com/bbc/bigscreen-player/pull/365/commits/2fb5b3c46f1f5767525478f8adf810011b9e5e80

src/playbackstrategy/msestrategy.js
```diff
@@ -448,13 +453,15 @@ function MSEStrategy(
   }
 
   function onCurrentTrackChanged(event) {
+    if (!isAudioDescribedAvailable()) return
+
     const mediaType = event.newMediaInfo.type
     DebugTool.info(
       `${mediaType} track changed.${
-        mediaType === "audio" ? (isBroadcastMixADEnabled() ? " BroadcastMixAD on." : " BroadcastMixAD off.") : ""
+        mediaType === "audio" ? (isAudioDescribedEnabled() ? " Audio Described on." : " Audio Described off.") : ""
```
eirikbjornr, line 461: Can this check is AD is available too for the logging?

terminal-overflow: As in, log out when it's available as well as when it's enabled?

src/playercomponent.js
terminal-overflow: Added in https://github.com/bbc/bigscreen-player/pull/365/commits/cde08da9b1974b3e3e9e63d0a9c5e498a04f8453.

src/playbackstrategy/msestrategy.js
ShiningTrapez: We return if it's not availible in the line directly before the cited block

eirikbjornr (approved):

PR #366, opened 2025-01-30
MSE BroadcastMixAD Support

Luke-Chatburn:
📺 What

Makes BroadcastMixAD available for MSE devices.

🛠 How

* Adds a set of new APIs that check for the availability, whether its enabled and ability to turn it on or off.
* Adds the ability to turn BroadcastMixAD on by default (if the track exists).
* Availability is determined from the audio properties in the manifest. Accessibility schema and the alternate role.
* Use dash.js to change the tracks between BroadcastMixAD and main audio.
* Adds unit tests.

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -455,9 +455,10 @@ function MSEStrategy(
   }
 
   function onCurrentTrackChanged(event) {
+    const mediaType = event.newMediaInfo.type
     DebugTool.info(
-      `${event.mediaType} track changed.${
-        event.mediaType === "audio" ? (isBroadcastMixADEnabled() ? " BroadcastMixAD on." : " BroadcastMixAD off.") : ""
+      `${mediaType} track changed.${
+        mediaType === "audio" ? (isBroadcastMixADEnabled() ? " BroadcastMixAD on." : " BroadcastMixAD off.") : ""
       }`
     )
     callBroadcastMixADCallbacks(isBroadcastMixADEnabled())
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -25,8 +25,8 @@ function MSEStrategy(
   mediaKind,
   playbackElement,
   isUHD,
-  enableBroadcastMixAD,
   customPlayerSettings,
+  enableBroadcastMixAD,
   callBroadcastMixADCallbacks
 ) {
   let mediaPlayer
@@ -315,11 +315,6 @@ function MSEStrategy(
       mediaPlayer.setMediaDuration(Number.MAX_SAFE_INTEGER)
     }
 
-    if (isBroadcastMixADAvailable() && enableBroadcastMixAD) {
-      setBroadcastMixADOn()
-      callBroadcastMixADCallbacks(true)
-    }
-
     emitPlayerInfo()
   }
 
@@ -460,7 +455,12 @@ function MSEStrategy(
   }
 
   function onTrackChangeRendered(event) {
-    DebugTool.info(`${event.mediaType} track changed. ${isBroadcastMixADEnabled() ? "AD on." : "AD off."}`)
+    DebugTool.info(
+      `${event.mediaType} track changed. ${
+        event.mediaType === "audio" ? (isBroadcastMixADEnabled() ? "BroadcastMixAD on." : "BroadcastMixAD off.") : ""
+      }`
+    )
+    callBroadcastMixADCallbacks(isBroadcastMixADEnabled())
   }
 
   function publishMediaState(mediaState) {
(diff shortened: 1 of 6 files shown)


src/playbackstrategy/msestrategy.js
```diff
@@ -444,6 +459,10 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
     Plugins.interface.onFragmentContentLengthMismatch(event)
   }
 
+  function onTrackChangeRendered(event) {
+    DebugTool.info(`${event.mediaType} track changed. ${isBroadcastMixADEnabled() ? "AD on." : "AD off."}`)
```
eirikbjornr, line 463: What's thoughts on calling `callBroadcastMixADCallbacks` here? IIRC Dash.js triggers this event switching to the first track too.

eirikbjornr: Semantics of calling that function here are different on second thought. Current impl calls the AD callbacks when a track switch is requested.

eirikbjornr: What's your opinion @edmorrish ?

src/playbackstrategy/msestrategy.js
```diff
@@ -305,6 +315,11 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
       mediaPlayer.setMediaDuration(Number.MAX_SAFE_INTEGER)
     }
 
+    if (isBroadcastMixADAvailable() && enableBroadcastMixAD) {
+      setBroadcastMixADOn()
```
eirikbjornr, line 319: What's the impact of triggering the track switch here vs using `setInitialMediaSettingsFor`? Do you see some segments of main audio loading before the switch?

terminal-overflow: I've read into this function more now and I believe that if the value isn't found, it default to another track instead of throwing an error. I'll give it a go and see if it works the same.

src/playbackstrategy/msestrategy.js
```diff
@@ -516,6 +535,14 @@ function MSEStrategy(mediaSources, windowType, mediaKind, playbackElement, isUHD
     mediaPlayer = MediaPlayer().create()
     mediaPlayer.updateSettings(dashSettings)
     mediaPlayer.initialize(mediaElement, null, true)
+
+    if (enableBroadcastMixAD) {
+      mediaPlayer.setInitialMediaSettingsFor("audio", {
```
eirikbjornr, line 540: What happens in the scenario where:

1. Viewer starts watching with `enableBroadcastMixAD: true`
2. Viewer disables broadcast mix AD
3. A CDN failover occurs

My thinking is: does setting the initial media settings mean the AD track is preferred on failovers?

src/playbackstrategy/msestrategy.test.js
```diff
@@ -572,6 +588,22 @@ describe("Media Source Extensions Playback Strategy", () => {
     })
   })
 
+  describe("Sets up mediaPlayer respecting enableBroadcastMixAD", () => {
+    it("sets initial audio track settings when enableBroadcastMixAD is true", () => {
+      setUpMSE(undefined, undefined, undefined, undefined, undefined, undefined, true)
+      mseStrategy.load(WindowTypes.STATIC, 10)
+
+      expect(mockDashInstance.setInitialMediaSettingsFor).toHaveBeenCalled()
```
eirikbjornr, line 596: Can you make this test more specific with `toHaveBeenCalledWith`?

src/playbackstrategy/msestrategy.test.js
```diff
@@ -1104,6 +1136,128 @@ describe("Media Source Extensions Playback Strategy", () => {
...
+    const mainTrack = { roles: ["main"] }
+
+    beforeEach(() => {
+      setUpMSE()
+      mseStrategy.load(WindowTypes.STATIC, 10)
+    })
+
+    describe("isBroadcastMixADAvailable()", () => {
+      it("returns true when there is a broadcastMixAD track", () => {
+        mockDashInstance.getTracksFor.mockReturnValueOnce([broadcastMixADtrack])
```
eirikbjornr, line 1154: Can you change these tests to return both tracks from the mock function? Just to check the selection logic doesn't pick the first track available

src/playbackstrategy/msestrategy.test.js
```diff
@@ -1104,6 +1136,128 @@ describe("Media Source Extensions Playback Strategy", () => {
...
+        mockDashInstance.getTracksFor.mockReturnValueOnce([mainTrack])
+
+        mseStrategy.setBroadcastMixADOn()
+
+        expect(mockDashInstance.setCurrentTrack).not.toHaveBeenCalled()
+      })
+    })
+
+    describe("onTrackChangeRendered", () => {
+      it("should ensure debugTool is called with true and callbacks are called with true when the current track is broadcastMixAD", () => {
```
eirikbjornr, line 1209: Is this test name correct? (Aside: Don't worry too much about unit testing info messages to the DebugTool – they don't make their way up to Chronicle so they're not used for data analysis)

eirikbjornr (changes requested):
Great work! My main bits of feedback are unit tests that could benefit from more specificity and a scenario I'm curious about

src/playbackstrategy/msestrategy.test.js
terminal-overflow: That's fine with me - I wasn't sure whether to include them. I've removed them now and updated the names.

src/playbackstrategy/msestrategy.js
terminal-overflow: After initial testing, it looks like the previous (not initial) setting is persisted on failover.

eirikbjornr (approved):
Excellent work!

PR #373, opened 2025-03-20
 Add Guards Around Setting `cached.currentTime`

terminal-overflow:
📺 What

Fixes an issue where `cached.currentTime` was not being set if `mediaElement.currentTime` was `0`. In situations where the `load` was called with no `initialPlaybackTime` and the `mediaElement.currentTime` was not naturally updated, the time would stay as `undefined`, causing the `readyHelper` to not call the success callback and causing BSP to be stuck in an unready state.

🛠 How

Add, correct and unify guards around the setting of `cached.currentTime` in MSE strategy to ensure the value is a number.

✅ Testing

Added a unit test covering the aforementioned scenario.

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -489,7 +489,9 @@ function MSEStrategy(
     if (mediaPlayer) {
       modifySource(cached.currentTime)
     } else {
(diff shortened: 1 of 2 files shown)


ShiningTrapez (approved):
Great Work :rocket:

PR #374, opened 2025-03-20
Use the Correct Name for Functions in MockBigScreenPlayer

terminal-overflow:
📺 What

Corrects a function name used in `MockBigScreenPlayer`.

🛠 How

Renames `setAudioDescribedEnabled` to `setAudioDescribed` to match the actual API.


ShiningTrapez (approved):
Good spot :eye:

PR #375, opened 2025-03-24
Add Correct Initial Audio Settings

terminal-overflow:
📺 What

Dash.js caches the last track used in local storage and selects the audio track based on selection priority and then max bitrate. This can result in the BroadcastMixAD (Audio Described) being selected against the users preference. This can manifest in two scenarios.

1. When the last track selected was BroadcastMixAD, the user turns off their preference out of stream and then the user plays a stream with a BroadcastMixAD track, the BroadcastMixAD track will be selected.
2. If the BroadcastMixAD track has a higher track selection priority in the manifest or has a higher max bitrate, the BroadcastMixAD audio track will be selected.

In both scenarios, the BroadcastMixAD track is selected when the users AD preference is off. This PR disables dash.js's media track caching and sets an initial audio preference.

🛠 How

- Disables `lastMediaSettingsCachingInfo`
- Sets initial audio settings to select `role: main` if Audio Described is not enabled at the start of stream.
- Adds relevant units tests.

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -42,6 +42,7 @@ function MSEStrategy(
           bufferTimeAtTopQuality: 12,
           bufferTimeAtTopQualityLongForm: 15,
         },
+        lastMediaSettingsCachingInfo: { enabled: false },
       },
     },
     customPlayerSettings
@@ -530,6 +531,10 @@ function MSEStrategy(
       mediaPlayer.setInitialMediaSettingsFor("audio", {
         role: "alternate",
         accessibility: { schemeIdUri: "urn:tva:metadata:cs:AudioPurposeCS:2007", value: "1" },
+      })
+    } else {
+      mediaPlayer.setInitialMediaSettingsFor("audio", {
+        role: "main",
       })
     }
 
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -527,16 +527,17 @@ function MSEStrategy(
     mediaPlayer.updateSettings(dashSettings)
     mediaPlayer.initialize(mediaElement, null)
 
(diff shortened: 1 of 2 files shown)


ShiningTrapez (dismissed):

ShiningTrapez:
Good Work!  :clap:

src/playbackstrategy/msestrategy.js
```diff
@@ -531,6 +532,10 @@ function MSEStrategy(
         role: "alternate",
         accessibility: { schemeIdUri: "urn:tva:metadata:cs:AudioPurposeCS:2007", value: "1" },
       })
+    } else {
+      mediaPlayer.setInitialMediaSettingsFor("audio", {
+        role: "main",
+      })
```
eirikbjornr, line 538: What's thoughts on folding it into one line? Like:

```suggestion
 mediaPlayer.setInitialMediaSettingsFor("audio", enableAD ? /* ad block */ : { role: "main" })
```

I'm thinking down the line that can be extended to a function to get the right settings block for the input value etc

eirikbjornr (approved):
Nice! One nitpick improvement so feel free to go ahead without it

src/playbackstrategy/msestrategy.js
ShiningTrapez: I think I may be biased here but I think we should do it

terminal-overflow: Haha no problem, will do that.

terminal-overflow: Done in https://github.com/bbc/bigscreen-player/pull/375/commits/0800f76a8a120104690e13377ebd0cf1f1d53536.

ShiningTrapez (dismissed):

eirikbjornr (dismissed):

eirikbjornr (approved):

PR #378, opened 2025-04-07
Add `target-latency` and `current-latency` to Chronicle (Update of #361)

ShiningTrapez:
📺 What

Adds MSE target latency and current latency reporting to the debug menu. Useful when debugging low-latency content, particularly on TV hardware where some decoder behaviours do not match what might be expected

🛠 How

Uses existing `DebugTool.dynamicMetric()` method to add these metrics each time a `timeupdate` event is called.

✅ Testing

Added underlying dash.js functions to the `mockDashInstance`.  No additional unit tests were added.

👀 See

New fields are seen below when the MSE strategy is chosen. The values` NaN` and 0 are seen when VOD content is played.

Screenshot 2024-10-30 at 17 06 22

The same changes are shown again, but using live low-latency content. Extra fields, along with the existing playback rate field, can be used to determine whether low-latency content is performing as expected.

Screenshot 2024-10-30 at 17 06 54

https://github.com/bbc/bigscreen-player/pull/361


ryanmccartney (dismissed):

ryanmccartney:
Updated to only show target and current latency values when playing back a live stream

src/playbackstrategy/msestrategy.js
```diff
@@ -197,6 +197,11 @@ function MSEStrategy(
   function onTimeUpdate() {
     DebugTool.updateElementTime(mediaElement.currentTime)
 
+    if (!isNaN(mediaPlayer.getCurrentLiveLatency())) {
+      DebugTool.dynamicMetric("current-latency", mediaPlayer.getCurrentLiveLatency())
+      DebugTool.dynamicMetric("target-latency", mediaPlayer.getTargetLiveDelay())
```
eirikbjornr, line 202: How often does this change over time? Concerned we'll fill the Chronicle logs with values so could these be changed to `staticMetric` instead?

eirikbjornr (approved):
LGTM! Left one comment on dynamic vs static metric

src/playbackstrategy/msestrategy.js
ryanmccartney: `getTargetLiveDelay` will be static but `getCurrentLiveLatency` will change over time. It's the important debug value to keep and eye on as it changes

ryanmccartney: I've moved `target-latency` to a `staticMetric`

eirikbjornr (approved):

PR #383, opened 2025-04-28
Add support for Dash.js rendered subtitles

ryanmccartney:
📺 What

Extends subtitle functionality by adding the ability to use Dash.js rendered subtitles if declared in an MPD rather than using the existing side-chain.

Allows the rendering of subtitles delivered in fragmented chunks, as used for low-latency playback.

It does not implement the current customisation options that are available in the existing side-chain as Dash.js provides no interface to access the customisation options present in the BBC fork of imscJS.

A currently open PR [imscJS #257](https://github.com/sandflow/imscJS/pull/257) would enable customisation on upstream imscJS and work on [Dash.js](https://github.com/bbc/rd-dash.js/tree/imsc-styling) [(imscJS](https://github.com/bbc/rd-dash.js/tree/imsc-styling)[ Styling](https://github.com/bbc/rd-dash.js/tree/imsc-styling) would allow this functionality to be added retrospectively.

🛠 How
By setting a new override - `embeddedSubtitles`, the side-chain is replaced by a new `embeddedSubtitles` object which enables text tracks in the MSE player

✅ Testing [Semi-optional]

Tests have been added to cover changes to the subtitle selector. In addition, coverage of the new dashsubtitles.js object is provided.

♿ Accessibility [optional]

Provides optional subtitles when using low-latency playback through Bigscreen player without modifications to the existing side-chain rendering.


nigelmegitt:
I'm missing something here: why can't we use a build of bbc/dash.js that includes the customisation functionality? It seems highly sub-optimal to disable it for low latency playback when most of the code exists.

Luke-Chatburn (dismissed):

markhouldridge (dismissed):
LGTM!  The only thing I spotted is two uses of `bsp_subtitles` id, but that can be sorted if we pickup the override subs element work, for the Short Form Video work 👍🏻

eirikbjornr:
> I'm missing something here: why can't we use a build of bbc/dash.js that includes the customisation functionality? It seems highly sub-optimal to disable it for low latency playback when most of the code exists.

@nigelmegitt we replace the open-source IMSC version included in Dash.js with the internal BBC version we pull in through BSP so the customisation functionality is enabled for low latency playback too

nigelmegitt:
> @nigelmegitt we replace the open-source IMSC version included in Dash.js with the internal BBC version we pull in through BSP so the customisation functionality is enabled for low latency playback too

Ah, thank you @eirikbjornr I didn't realise that. OK, that's fine then.

src/playbackstrategy/msestrategy.js
```diff
@@ -615,6 +629,11 @@ function MSEStrategy(
     const anchor = buildSourceAnchor(presentationTimeInSeconds)
 
     mediaPlayer.attachSource(`${source}${anchor}`)
+
+    if (embeddedSubs) {
```
eirikbjornr, line 633: What's the reason to attach the rendering div after attaching the source? My assumption is you want the subtitles config set up before the stream is loaded. Does it need to be reattached for each new source?

ShiningTrapez: We just did it like this so new stuff was at the bottom

eirikbjornr: Fair yea moving it to `setupMediaPlayer` feels like a substantive change that'd need a quick re-check so I understand why you don't want to do that – IMO better to move it above the `attachSource` call inside `modifySource` though

eirikbjornr (approved):

eirikbjornr:
Amazing work! You've perservered through some gnarly bugs

PR #384, opened 2025-05-06
Use full entrypoint of dash.js, allow protection data to be used

jlks:
📺 What

Take the full dash.js entrypoint vs `index_mediaplayerOnly`, pass through dash.js protection data.

🛠 How

- Update the dash.js entrypoint. Use the index vs the built `main` package entrypoint to allow consuming applications to decide what to bundle.
- Use the same shape protection data as dash.js within the optional `media.protectionData`. This is exposed via the types.
- Expose the protection data within mediaSources with the `currentProtectionData` function.
- Set protection data on the dash.js media player.

### DIFF
--- a/src/mediasources.ts
+++ b/src/mediasources.ts
@@ -1,4 +1,4 @@
-import type { CaptionsConnection, Connection, MediaDescriptor } from "./types"
+import type { CaptionsConnection, Connection, MediaDescriptor, MediaProtectionData } from "./types"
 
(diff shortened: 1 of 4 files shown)


ShiningTrapez (approved):
How much does taking the full bundle of Dash increase the library size?

jlks:
> How much does taking the full bundle of Dash increase the library size?

Roughly 300kB

eirikbjornr (approved):

PR #390, opened 2025-06-02
Initialise before setting protection data in MSE Strategy

jlks:
📺 What

Initialise before setting protection data (as in dash.js examples). Add extra info logging.

🛠 How

- Move `initialize` call
- Info log `[Protection` prefixed messages to cover calls through to ProtectionController/ProtectionModel.


src/playbackstrategy/msestrategy.js
```diff
@@ -506,7 +507,9 @@ function MSEStrategy(
   }
 
   function onDebugLog(event) {
-    DebugTool.debug(event.message)
```
eirikbjornr, line 509: Can we retain this with an `if-else`? So we've got Dash.js logs on `DebugTool.debug`

ShiningTrapez (dismissed):

eirikbjornr (changes requested):
Small change I'd really like

PR #391, opened 2025-06-09
Support setting and getting of playback bitrate

markhouldridge:
📺 What

Changes to interface to support the setting of a bitrate, given a media kind (Audio/video), minimum and maximum bitrate for the selection of a representation within the bitrate range.

Provides an interface for returning the current bitrate setting.

🛠 How

Two new functions, `setConstrainedBitrateInKbps` and `getPlaybackBitrate` for setting and getting the playback bitrate respectively.

### DIFF
--- a/src/bigscreenplayer.js
+++ b/src/bigscreenplayer.js
@@ -472,6 +472,24 @@ function BigscreenPlayer() {
     getPlaybackRate: () => playerComponent && playerComponent.getPlaybackRate(),
 
     /**
+     * Set constrained bitrate given a min/max range and mediakind.
+     */
+    setConstrainedBitrateInKbps: (mediaKind, minBitrate, maxBitrate) => {
+      if (playerComponent) {
+        playerComponent.setConstrainedBitrateInKbps(mediaKind, minBitrate, maxBitrate)
+      }
+    },
+
+    /**
+     * Returns current playback bitrate for media kind.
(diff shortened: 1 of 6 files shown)


src/bigscreenplayer.js
```diff
@@ -471,6 +471,24 @@ function BigscreenPlayer() {
      */
     getPlaybackRate: () => playerComponent && playerComponent.getPlaybackRate(),
 
+    /**
+     * Set constrained bitrate given a min/max range and mediakind.
+     */
+    setConstrainedBitrateInKbps: (mediaKind, minBitrate, maxBitrate) => {
```
ShiningTrapez, line 477: `setPlaybackBitrate` for Symmetry? The detail that it's in kbps can be in the Doc comment IMO (Also that detail being in the function name kinda implies, for example, `setConstrainedBitrateInMbps` on the API)

ShiningTrapez (approved):
Just one comment :D

src/bigscreenplayer.js
markhouldridge: Thank you both for the review feedback - I've changed the interface for setting bitrate to `setBitrateConstraint` and params to `minBitrateKbps` and `maxBitrateKbps` which I agree, is clearer.   I've gone with `constraint` wording as this hints of a range applied.

PR #393, opened 2025-07-03
Fix type errors on video only content

Luke-Chatburn:
📺 What

Ensure no type errors on video only content

🛠 How

Check for video/audio tracks in the manifest.

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -327,13 +327,15 @@ function MSEStrategy(
       mediaPlayer.setMediaDuration(Number.MAX_SAFE_INTEGER)
     }
 
-    if (mediaKind === MediaKinds.VIDEO) {
+    if (mediaPlayer.getActiveStream()?.getHasVideoTrack()) {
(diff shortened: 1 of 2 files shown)


src/playbackstrategy/msestrategy.js
```diff
@@ -376,9 +376,11 @@ function MSEStrategy(
   }
 
   function dispatchMaxQualityChangeForKind(kind) {
-    const { qualityIndex, bitrate: bitrateInBps } = mediaPlayer.getTopBitrateInfoFor(kind)
+    const { qualityIndex, bitrate: bitrateInBps } = mediaPlayer.getTopBitrateInfoFor(kind) ?? {}
 
-    DebugTool.dynamicMetric(`${kind}-max-quality`, [qualityIndex, bitrateInBps])
+    if (qualityIndex && bitrateInBps) {
```
jlks, line 381: Is it possible we get a similar issue with `dispatchDownloadQualityChangeForKind`, `emitPlayerInfo` and `onStreamInitialised` above too? they all utilise similar function calls I think.

eirikbjornr (approved):

PR #400, opened 2025-08-20
Introduce Abort Initialisation After TearDown

terminal-overflow:
📺 What

Introduces the concept of aborting initialisation if Bigscreen Player has already been torn down.

This prevents the issue where, if BSP is torn down during initialisation, `TypeError: Cannot read properties of null (reading 'manifestType')` or `Cannot read properties of undefined (reading 'time')` is thrown, depending on where `mediaSources.time().manifestType` is called. These errors can be caused by `tearDown()` tearing down and de-referencing `mediaSources` **whilst** BSP is in the process of initialising.

🛠 How

- Introduce an `AbortSignal` which keeps the state of if BSP has been torn down. Also provides a `throwIfAborted` error which throws an `AbortError`.
  - The `AbortError` has a custom name and message.
- Call `signal.abort()` on invocation of `tearDown()`.
- Add unit tests for `abortutils`.
- Add integration tests to `bigscreenplayer` and `playercomponent`.

✅ Testing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "bigscreen-player",
-  "version": "10.10.2",
+  "version": "10.11.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "bigscreen-player",
-      "version": "10.10.2",
+      "version": "10.11.1",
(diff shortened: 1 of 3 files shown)


ShiningTrapez (approved):
LGTM 🚀

PR #406, opened 2025-09-29
Add video fragment response headers to the debug view

jlks:
📺 What

For MSE browsers, show the headers in `debug.fragmentResponseHeaders ` as static fields in the debug tool overlay.

🛠 How
- Add new optional `debug` object to the initialisation data.
- For each header in `fragmentResponseHeaders` display the value from the latest video fragment response.

### DIFF
--- a/src/playbackstrategy/msestrategy.js
+++ b/src/playbackstrategy/msestrategy.js
@@ -22,7 +22,8 @@ function MSEStrategy(
   playbackElement,
   _isUHD = false,
   customPlayerSettings = {},
-  audioDescribedOpts = {}
+  audioDescribedOpts = {},
+  debugSettings = {}
 ) {
   const audioDescribed = { callback: undefined, enable: false, ...audioDescribedOpts }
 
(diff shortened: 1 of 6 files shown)


src/playbackstrategy/msestrategy.js
```diff
@@ -524,6 +526,19 @@ function MSEStrategy(
...
   }
 
+  function onFragmentLoaded() {
+    if (debugSettings?.fragmentResponseHeaders) {
+      debugSettings.fragmentResponseHeaders.forEach((responseHeader) => {
+        const responseHeaderValue = mediaPlayer
+          .getDashMetrics()
+          .getLatestFragmentRequestHeaderValueByID("video", responseHeader)
+        if (responseHeaderValue) {
+          DebugTool.staticMetric(responseHeader, responseHeaderValue)
```
eirikbjornr, line 536: We need some way to either:
- filter out segment response header metrics from the Chronicle payload from `getChronicle`, or
- add segment response headers to the types so they're properly validated

ATM this will break the types and in turn telemetry hooking off that

eirikbjornr (commented):
Got some concerns around the typing. Didn't make the connection here before or I'd raised it sooner

PR #426, opened 2026-04-22
BADGER-364: Remove UHD overrides

JoeBBC:
📺 What

Removes legacy override which disables sentinels when consuming UHD content.

🛠 How

- Removes the config check, explicitly set `disableSentinels` to false
- Removes relevant unit test


eirikbjornr (approved):

stomeska:
> 📺 What
> 
> Removes legacy override which disables sentinels when consuming UHD content.
> 
> 🛠 How
> 
> - Removes the config check, explicitly set `disableSentinels` to false
> - Removes relevant unit test

PR #428, opened 2026-09-15
Add npmrc with min release age

eirikbjornr:
📺 What

Improve npmrc settings for dependency security.

🛠 How

- Husky is no longer installed on a postinstall hook, it's a separate command.


markhouldridge (approved):
lgtm 👍🏻

